PR #13, opened 2016-08-21
build.sh: added missing ref to Generated.Custom.cs, format commands

andyli:
Reference to `Generated.Custom.cs` is missed, leads to build error:
```
ClangSharpPInvokeGenerator/Program.cs(106,50): error CS0117: `ClangSharp.clang' does not contain a definition for `parseTranslationUnit2'
```

Formatted the commands into multiple lines for readability.

Permission of build.sh is changed from 644 → 755.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,6 +7,29 @@ if [ -z "$2" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


mjsabby:
LGTM. Thanks!

PR #15, opened 2017-03-19
Properly handle CXTypeKind.CXType_Elaborated

stanionascu:
Otherwise generated code resolves the types as UnknownType(s)


mjsabby:
Thanks @stanionascu

PR #16, opened 2017-04-27
Crash fix with const char* return type, Clang 4.0 support

blue3k:
Fix for crash while calling functions with const char* return.  …
 - C# trying to delete string buffer allocated from std heap.
 - It will create wrapper function and use Marshal.PtrToStringAnsi to convert

Clang 4.0 support
 - Elaborated type support

### DIFF
--- a/ClangSharp.csproj
+++ b/ClangSharp.csproj
@@ -29,6 +29,26 @@
     <WarningLevel>4</WarningLevel>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64'">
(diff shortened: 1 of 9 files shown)


mjsabby:
Hey @blue3k while I review your changes, can you squash your changes into a single commit?

blue3k:
I thought you might want to review them separately.
I squashed them into one. CL 6b3bebc28bd39d4a772fc6ea5bae212aa635a16d
Can you check that again?

SolalPirelli:
@blue3k Do you think this is still needed? In my part of #21 I changed the marshaler for strings, FWIW.

tannergooding:
Going to close this as the repo has undergone a significant rewrite and it should now have correct marshaling behavior when dealing with strings.

It is also now support clang 8.0.0.

PR #18, opened 2017-11-08
Upgrade to Clang 5.0.0

amaitland:
I've run the Generator against `Clang 5.0.0` and this is the result

- Removed `index_setClientEntity` and `index_getClientEntity` from generated code, same as https://github.com/Microsoft/ClangSharp/pull/16. If this is incorrect then pointers welcome.
- Added additional check for `CXTypeKind.CXType_Elaborated` as per https://github.com/Microsoft/ClangSharp/pull/16 this can be easily reverted if required.

The `Generated.cs` and `Generated.Custom.cs` files contained within `ClangSharpPInvokeGenerator` are different to those in root, is there a reason they're different? I've updated them so they're now the same, this can easily be reverted.

I'm interested in a `Clang 5.0.0` release and I'm happy to contribute where possible. I've kept the three distinct commits as it's now very easy to `Squash Merge` using the `GitHub` interface and I personally feel it makes the changes easy to review. Feedback welcome.

### DIFF
--- a/ClangSharpPInvokeGenerator/FunctionVisitor.cs
+++ b/ClangSharpPInvokeGenerator/FunctionVisitor.cs
@@ -13,12 +13,21 @@
 
         private readonly string prefixStrip;
 
-        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip)
+        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip, string[] execludeFunctionsArray)
         {
             this.prefixStrip = prefixStrip;
             this.tw = tw;
             this.tw.WriteLine("        private const string libraryPath = \"" + libraryPath + "\";");
             this.tw.WriteLine();
+
+            if(execludeFunctionsArray != null)
+            {
+                //Fir all the functions we're excluding add them to the visitedFunctions set so they're ignored
+                foreach(var func in execludeFunctionsArray)
+                {
+                    visitedFunctions.Add(func);
+                }
+            }
         }
 
         public CXChildVisitResult Visit(CXCursor cursor, CXCursor parent, IntPtr data)
--- a/ClangSharpPInvokeGenerator/FunctionVisitor.cs
+++ b/ClangSharpPInvokeGenerator/FunctionVisitor.cs
@@ -13,17 +13,17 @@
 
         private readonly string prefixStrip;
 
-        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip, string[] execludeFunctionsArray)
+        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip, string[] excludeFunctionsArray)
         {
             this.prefixStrip = prefixStrip;
             this.tw = tw;
             this.tw.WriteLine("        private const string libraryPath = \"" + libraryPath + "\";");
             this.tw.WriteLine();
 
-            if(execludeFunctionsArray != null)
+            if(excludeFunctionsArray != null)
             {
-                //Fir all the functions we're excluding add them to the visitedFunctions set so they're ignored
-                foreach(var func in execludeFunctionsArray)
+                //For all the functions we're excluding add them to the visitedFunctions set so they're ignored
+                foreach(var func in excludeFunctionsArray)
                 {
                     visitedFunctions.Add(func);
                 }
--- a/ClangSharpPInvokeGenerator/Program.cs
+++ b/ClangSharpPInvokeGenerator/Program.cs
@@ -4,11 +4,19 @@
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Runtime.InteropServices;
     using System.Text.RegularExpressions;
     using ClangSharp;
 
     public class Program
     {
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern IntPtr LoadLibraryEx(string lpFileName, IntPtr hReservedNull, uint dwFlags);
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        private static extern bool FreeLibrary(IntPtr hModule);
+
         public static void Main(string[] args)
         {
             Regex re = new Regex(@"(?<switch>-{1,2}\S*)(?:[=:]?|\s+)(?<value>[^-\s].*?)?(?=\s+[-]|$)");
@@ -23,6 +31,8 @@
             string libraryPath = string.Empty;
             string prefixStrip = string.Empty;
             string methodClassName = "Methods";
+            string libClangPath = string.Empty;
+            IntPtr libClangHandle = IntPtr.Zero;
 
             foreach (KeyValuePair<string, string> match in matches)
             {
@@ -60,6 +70,11 @@
                 {
                     methodClassName = match.Value;
                 }
+
+                if (string.Equals(match.Key, "--c") || string.Equals(match.Key, "--clangPath"))
+                {
+                    libClangPath = match.Value;
+                }
             }
 
             var errorList = new List<string>();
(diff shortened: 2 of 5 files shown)


amaitland:
Happy to sign the `cla` when I'm next back in front of a `PC`.

amaitland:
> Removed index_setClientEntity and index_getClientEntity from generated code, same as #16. If this is incorrect then pointers welcome.

Looked into this further and appears the generated ones should be excluded, along with a number of other generated functions, these are just the ones that clashed. Added `--excludeFunctions` command line argument and updated the batch file to exclude what are hopefully all the custom generated functions.

> Added additional check for CXTypeKind.CXType_Elaborated as per #16 this can be easily reverted if required.

Debugging this a little more, the output appears to be the same, I think it's a minor improvement when debugging, so keeping it seems reasonable.

I've added `--clangPath` command line argument to make life easier when debugging, adding the ability to load `libclang` from a different directory.

amaitland:
> Debugging this a little more, the output appears to be the same, I think it's a minor improvement when debugging, so keeping it seems reasonable.

Changed my mind, removed as it's unnecessary

ClangSharpPInvokeGenerator/FunctionVisitor.cs
```diff
@@ -13,12 +13,21 @@ internal sealed class FunctionVisitor : ICXCursorVisitor
 
         private readonly string prefixStrip;
 
-        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip)
+        public FunctionVisitor(TextWriter tw, string libraryPath, string prefixStrip, string[] execludeFunctionsArray)
```
mjsabby: spelling

ClangSharpPInvokeGenerator/FunctionVisitor.cs
```diff
@@ -13,12 +13,21 @@ internal sealed class FunctionVisitor : ICXCursorVisitor
...
         {
             this.prefixStrip = prefixStrip;
             this.tw = tw;
             this.tw.WriteLine("        private const string libraryPath = \"" + libraryPath + "\";");
             this.tw.WriteLine();
+
+            if(execludeFunctionsArray != null)
+            {
+                //Fir all the functions we're excluding add them to the visitedFunctions set so they're ignored
```
mjsabby: spelling: Fire.

ClangSharpPInvokeGenerator/Generated.Custom.cs
```diff
@@ -117,30 +117,38 @@ public static IntPtr codeCompleteAt(CXTranslationUnit @TU, string @complete_file
         public static int indexSourceFile(CXIndexAction @param0, CXClientData @client_data, IndexerCallbacks[] @index_callbacks, uint @index_callbacks_size, uint @index_options, string @source_filename, string[] @command_line_args, int @num_command_line_args, CXUnsavedFile[] @unsaved_files, uint @num_unsaved_files, out CXTranslationUnit @out_TU, uint @TU_options)
         {
             var arr = new _CXUnsavedFile[unsaved_files.Length];
+            IntPtr result = IntPtr.Zero;
 
             try
             {
                 BeginCXUnsavedFileMarshal(ref arr, ref unsaved_files);
-                return indexSourceFile(param0, client_data, index_callbacks, index_callbacks_size, index_options, source_filename, command_line_args, num_command_line_args, arr, num_unsaved_files, out out_TU, TU_options);
+                BeginIndexerCallbacksMarshal(out result, ref index_callbacks);
```
mjsabby: I'm assuming this was a bug in the original code and you're having to marshal this correctly?

ClangSharpPInvokeGenerator/Generated.Custom.cs
```diff
@@ -117,30 +117,38 @@ public static IntPtr codeCompleteAt(CXTranslationUnit @TU, string @complete_file
...
         {
             var arr = new _CXUnsavedFile[unsaved_files.Length];
+            IntPtr result = IntPtr.Zero;
 
             try
             {
                 BeginCXUnsavedFileMarshal(ref arr, ref unsaved_files);
-                return indexSourceFileFullArgv(param0, client_data, index_callbacks, index_callbacks_size, index_options, source_filename, command_line_args, num_command_line_args, arr, num_unsaved_files, out out_TU, TU_options);
+                BeginIndexerCallbacksMarshal(out result, ref index_callbacks);
+
```
mjsabby: extra linebreak.

ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -85,11 +107,24 @@ public static void Main(string[] args)
...
                 foreach (var error in errorList)
                 {
                     Console.WriteLine(error);
                 }
+
+                //Terminate execution as command line arguments are missing
+                Environment.Exit(-1);
+            }
+
+            if(!string.IsNullOrEmpty(libClangPath) && File.Exists(libClangPath) && (libClangPath.EndsWith(".dll") || libClangPath.EndsWith(".so")))
```
mjsabby: why do you need this?

ClangSharpPInvokeGenerator/Generated.cs
```diff
@@ -1911,13 +1911,13 @@ public static partial class clang
         public static extern CXSourceRange getTokenExtent(CXTranslationUnit @param0, CXToken @param1);
 
         [DllImport(libraryPath, EntryPoint = "clang_tokenize", CallingConvention = CallingConvention.Cdecl)]
-        public static extern void tokenize(CXTranslationUnit @TU, CXSourceRange @Range, out CXToken[] @Tokens, out uint @NumTokens);
+        public static extern void tokenize(CXTranslationUnit @TU, CXSourceRange @Range, out IntPtr @Tokens, out uint @NumTokens);
 
         [DllImport(libraryPath, EntryPoint = "clang_annotateTokens", CallingConvention = CallingConvention.Cdecl)]
-        public static extern void annotateTokens(CXTranslationUnit @TU, CXToken[] @Tokens, uint @NumTokens, out CXCursor @Cursors);
+        public static extern void annotateTokens(CXTranslationUnit @TU, out CXToken @Tokens, uint @NumTokens, out CXCursor @Cursors);
```
mjsabby: any reason why you've removed the [] and made it out?

ClangSharpPInvokeGenerator/Generated.Custom.cs
amaitland: This whole commit is simply the result of copying `Generated.cs` and `Generated.Custom.cs` from the root over the top of  `ClangSharpPInvokeGenerator\Generated.cs` and `ClangSharpPInvokeGenerator\Generated.Custom.cs`.

Just making them consistent. Happy to revert this whole commit if you'd prefer?

ClangSharpPInvokeGenerator/Generated.cs
amaitland: Direct copy of  https://github.com/Microsoft/ClangSharp/blob/master/Generated.cs#L1917

ClangSharpPInvokeGenerator/FunctionVisitor.cs
amaitland: Fixed, thx

ClangSharpPInvokeGenerator/FunctionVisitor.cs
amaitland: Fixed, thx

ClangSharpPInvokeGenerator/Program.cs
amaitland: Which part? Would you prefer a console message is written if `libclang` doesn't exist? Or are you referring to the feature in general? 

I admit the code it's self is a little on the defensive side, makes sure that the file exists, that it's a `dll` or `so`, all seem reasonable checks, though looking back maybe writing a message to the console and terminating execution would be better?

ClangSharpPInvokeGenerator/Generated.Custom.cs
amaitland: Thinking about this more, this is probably my mistake, I probably synced the files in the incorrect direction....

amaitland: Maybe the two of the file should be removed from `Git` and copied as part of the build/post build process? (Assuming `VS` plays nice with the missing files)

Unless I'm missing something and there's a reason they're different.

amaitland:
Removed the unnecessary commit that synced the `Generated.cs` and `Generated.Custom.cs` files, was incorrect.

benpye:
@mjsabby Is there any reason this hasn't been looked at again? Looking to use ClangSharp but binding to such an old version as 3.x is a blocker. Even this is somewhat out of date now with 6.0.0 released.

amaitland:
Closing in favor of #21

PR #19, opened 2017-11-08
Add return MarshalAs attribute

amaitland:
For params the `MarshalAs` attribute is applied, this applies it for the `return param` as well.

Currently only for strings

### DIFF
--- a/ClangSharpPInvokeGenerator/Extensions.cs
+++ b/ClangSharpPInvokeGenerator/Extensions.cs
@@ -20,7 +20,10 @@
                 switch (pointee.kind)
                 {
                     case CXTypeKind.CXType_Char_S:
+                    case CXTypeKind.CXType_WChar:
+                    {
                         return true;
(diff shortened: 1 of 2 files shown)


ClangSharpPInvokeGenerator/Extensions.cs
```diff
@@ -117,8 +120,29 @@ public static void WriteFunctionInfoHelper(CXCursor cursor, TextWriter tw, strin
             var functionType = clang.getCursorType(cursor);
             var functionName = clang.getCursorSpelling(cursor).ToString();
             var resultType = clang.getCursorResultType(cursor);
+
+
```
mjsabby: nit: extra line.

amaitland:
> Just a nit on an extra line, otherwise LGTM.

Done 👍

amaitland:
With #21 it would appear this is no longer required, closing.

PR #32, opened 2019-04-24
 Adding more extensions to expose APIs in an object oriented manner.

tannergooding:
This is a continuation of https://github.com/Microsoft/ClangSharp/pull/31. It ports a good number of additional APIs and adds a simple test covering the new OOP extensions. It also moves ClangSharpPInvokeGenerator to use the OOP extensions exclusively (both to clean it up and to validate things are exposed correctly).

After this is merged, the following methods still need to be made available: [todo.txt](https://github.com/Microsoft/ClangSharp/files/3110051/todo.txt). I'm going to log a tracking issue for them, however, as I think they can be handled separately and aren't going to be targeted as much.

### DIFF
--- a/ClangSharp.Test/TranslationUnit.cs
+++ b/ClangSharp.Test/TranslationUnit.cs
@@ -36,5 +36,35 @@ namespace ClangSharp.Test
                 Directory.Delete(dir, true);
             }
         }
+
+        [Theory]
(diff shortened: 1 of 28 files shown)


mjsabby:
@tannergooding I've made you an admin, so after we review each other's work we can both commit.

PR #40, opened 2019-04-29
More improvements to the P/Invoke Binding Generator

tannergooding:
This does some additional cleanup of the P/Invoke Binding Generator and allowed https://github.com/Microsoft/LLVMSharp/pull/93 to be regenerated and fixed to work end to end.

The primary changes are:
* Start carrying around a "ConfigurationOptions" for the writer
* Change node traversal to be a two pass process, rather than a 4 pass process
  * This has an added "benefit" that the output more closely mirrors the input files in terms of ordering
* Some small fixes around typedefs and parameters that weren't being handled
* Basic support for generating unsafe code or decide to not generate wrapper types


ClangSharp/Generated.cs
```diff
@@ -3,294 +3,6 @@ namespace ClangSharp
     using System;
     using System.Runtime.InteropServices;
 
-    public partial struct CXTargetInfoImpl
```
tannergooding: The changes here are a side effect of doing two passes. I validated manually there was zero diff without that commit.

ClangSharpPInvokeGenerator/CXCursorExtensions.cs
```diff
@@ -1,329 +0,0 @@
-﻿using System.Diagnostics;
-using ClangSharp;
-
-namespace ClangSharpPInvokeGenerator
```
tannergooding: These all got moved into `CursorWriter` so they could depend on the `ConfigurationOptions`

ClangSharpPInvokeGenerator/Properties/launchSettings.json
```diff
@@ -2,17 +2,11 @@
   "profiles": {
     "Regenerate ClangSharp - Windows": {
       "commandName": "Project",
-      "commandLineArgs": "--m clang --p clang_ --namespace ClangSharp --output ../../../../ClangSharp/Generated.cs --libraryPath $(LibClangName) --include $(LLVMIncludePath) --file $(LLVMIncludePath)/clang-c/Index.h --file $(LLVMIncludePath)/clang-c/CXString.h --file $(LLVMIncludePath)/clang-c/Documentation.h --file $(LLVMIncludePath)/clang-c/CXErrorCode.h --file $(LLVMIncludePath)/clang-c/BuildSystem.h --file $(LLVMIncludePath)/clang-c/CXCompilationDatabase.h --excludeFunctions clang_index_getClientEntity,clang_index_setClientEntity,clang_createTranslationUnitFromSourceFile,clang_parseTranslationUnit,clang_parseTranslationUnit2,clang_parseTranslationUnit2FullArgv,clang_reparseTranslationUnit,clang_codeCompleteAt,clang_indexSourceFile,clang_indexSourceFileFullArgv,clang_disposeSourceRangeList,clang_annotateTokens,clang_annotateTokens,clang_disposeTokens,clang_tokenize,clang_disposeStringSet,clang_getCompletionParent,clang_disposeCXPlatformAvailability,clang_getCursorPlatformAvailability",
-      "environmentVariables": {
-        "LLVMIncludePath": "C:\\Program Files\\LLVM\\include"
```
tannergooding: These weren't needed as the actual setting was coming from the csproj file, which is probably better.

PR #43, opened 2019-05-09
Support multi file generation in the Binding generator

tannergooding:
This resolves https://github.com/microsoft/ClangSharp/issues/30

This updates the ClangSharpPInvokeGenerator to support emitting one file per type (and one file for all methods).

It also updates the generator to only emit required usings and to not emit a class to hold methods if the header file contains no methods.


tannergooding:
This allows a single pass per translation unit.

This has a side effect of always emitting two files, even if `multi-file` is not specified (one for the types and one for the methods). I could probably fix this to continue being one file, but it probably needs more thought on how I can do that cleanly.

I did this in a couple stages to validate that nothing regressed in the overall generation story. Notably, the first commit adds support for multi-file generation but just regenerates the code into two files. The second commit then regenerates things split up.

tannergooding:
Overall, having things split into one file per type allows easier review of changes and makes the output deterministic, rather than dependent on various parse orderings (such as passing in `y.h` before `x.h`).

ClangSharpPInvokeGenerator/OutputBuilder.cs
```diff
@@ -0,0 +1,140 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace ClangSharpPInvokeGenerator
+{
+    internal sealed class OutputBuilder : IDisposable
```
tannergooding: This could also allow some other interesting configuration options. Such as sorting types or methods before writing them out to the file (which would allow determinism for single-file generation and for the method class file).

ClangSharp/Generated/clang.cs
```diff
@@ -2533,6 +1088,5 @@ public static partial class clang
 
         [DllImport(libraryPath, EntryPoint = "clang_CompileCommand_getMappedSourceContent", CallingConvention = CallingConvention.Cdecl)]
         public static extern CXString CompileCommand_getMappedSourceContent(CXCompileCommand param0, uint I);
-
```
tannergooding: The `OutputBuilder` also makes it easier to handle these trailing newlines 🎉

PR #45, opened 2019-05-11
Adding support for parsing expressions attached to EnumConstantDecl

tannergooding:
This just updates the binding generator to traverse expressions attached to enum constant declarations. This means that the generated code will more closely match the source files.


ClangSharp/Generated/CXAvailabilityKind.cs
```diff
@@ -2,9 +2,9 @@ namespace ClangSharp
 {
     public enum CXAvailabilityKind
     {
-        CXAvailability_Available = 0,
-        CXAvailability_Deprecated = 1,
-        CXAvailability_NotAvailable = 2,
-        CXAvailability_NotAccessible = 3,
+        CXAvailability_Available,
```
tannergooding: The header file doesn't explicitly set the values for most enums and we don't need to either.

ClangSharp/Generated/CXCallingConv.cs
```diff
@@ -13,7 +13,7 @@ public enum CXCallingConv
         CXCallingConv_X86RegCall = 8,
         CXCallingConv_IntelOclBicc = 9,
         CXCallingConv_Win64 = 10,
-        CXCallingConv_X86_64Win64 = 10,
+        CXCallingConv_X86_64Win64 = CXCallingConv_Win64,
```
tannergooding: An example of traversing a declaration reference expression

ClangSharp/Generated/CXCodeComplete_Flags.cs
```diff
@@ -2,10 +2,10 @@ namespace ClangSharp
 {
     public enum CXCodeComplete_Flags
     {
-        CXCodeComplete_IncludeMacros = 1,
-        CXCodeComplete_IncludeCodePatterns = 2,
-        CXCodeComplete_IncludeBriefComments = 4,
-        CXCodeComplete_SkipPreamble = 8,
-        CXCodeComplete_IncludeCompletionsWithFixIts = 16,
+        CXCodeComplete_IncludeMacros = 0x01,
```
tannergooding: An example of traversing an integer literal expression

ClangSharp/Generated/CXCompletionContext.cs
```diff
@@ -3,29 +3,29 @@ namespace ClangSharp
...
-        CXCompletionContext_ObjCProtocol = 32768,
-        CXCompletionContext_ObjCCategory = 65536,
-        CXCompletionContext_ObjCInstanceMessage = 131072,
-        CXCompletionContext_ObjCClassMessage = 262144,
-        CXCompletionContext_ObjCSelectorName = 524288,
-        CXCompletionContext_MacroName = 1048576,
-        CXCompletionContext_NaturalLanguage = 2097152,
-        CXCompletionContext_IncludedFile = 4194304,
-        CXCompletionContext_Unknown = 8388607,
+        CXCompletionContext_AnyType = 1 << 0,
```
tannergooding: An example of traversing a binary operator expression

ClangSharp/Generated/CXCompletionContext.cs
```diff
@@ -3,29 +3,29 @@ namespace ClangSharp
...
+        CXCompletionContext_ObjCInterface = 1 << 14,
+        CXCompletionContext_ObjCProtocol = 1 << 15,
+        CXCompletionContext_ObjCCategory = 1 << 16,
+        CXCompletionContext_ObjCInstanceMessage = 1 << 17,
+        CXCompletionContext_ObjCClassMessage = 1 << 18,
+        CXCompletionContext_ObjCSelectorName = 1 << 19,
+        CXCompletionContext_MacroName = 1 << 20,
+        CXCompletionContext_NaturalLanguage = 1 << 21,
+        CXCompletionContext_IncludedFile = 1 << 22,
+        CXCompletionContext_Unknown = ((1 << 23) - 1),
```
tannergooding: even works with parentheses

ClangSharp/Generated/CXGlobalOptFlags.cs
```diff
@@ -2,9 +2,9 @@ namespace ClangSharp
...
     public enum CXGlobalOptFlags
     {
-        CXGlobalOpt_None = 0,
-        CXGlobalOpt_ThreadBackgroundPriorityForIndexing = 1,
-        CXGlobalOpt_ThreadBackgroundPriorityForEditing = 2,
-        CXGlobalOpt_ThreadBackgroundPriorityForAll = 3,
+        CXGlobalOpt_None = 0x0,
+        CXGlobalOpt_ThreadBackgroundPriorityForIndexing = 0x1,
+        CXGlobalOpt_ThreadBackgroundPriorityForEditing = 0x2,
+        CXGlobalOpt_ThreadBackgroundPriorityForAll = CXGlobalOpt_ThreadBackgroundPriorityForIndexing | CXGlobalOpt_ThreadBackgroundPriorityForEditing,
```
tannergooding: and combining multiple declaration references

ClangSharp/Generated/CXLanguageKind.cs
```diff
@@ -3,8 +3,8 @@ namespace ClangSharp
     public enum CXLanguageKind
     {
         CXLanguage_Invalid = 0,
```
tannergooding: An example of the native header explicitly setting the value for one entry, but not the subsequent declarations

ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -89,12 +89,17 @@ public static int Run(InvocationContext context)
             arr = arr.Concat(defines.Select(x => "-D" + x)).ToArray();
             arr = arr.Concat(additionalArgs).ToArray();
 
+            var translationFlags = CXTranslationUnit_Flags.CXTranslationUnit_None;
```
tannergooding: Also updated this to start passing in some flags. Namely, not processing function bodies currently and including implicit attribute information (which ensures we aren't missing anything if the user specifies additional command line options).

ClangSharpPInvokeGenerator/CursorVisitor.cs
```diff
@@ -258,7 +268,7 @@ public CXChildVisitResult VisitTranslationUnit(CXCursor cursor, CXCursor parent,
         {
             Debug.Assert(parent.Kind == CXCursorKind.CXCursor_TranslationUnit);
 
-            if (cursor.Location.IsInSystemHeader || !cursor.Location.IsFromMainFile)
+            if (!cursor.Location.IsFromMainFile)
```
tannergooding: Changed this to just check if we aren't in the main file; which ultimately allows users to generate bindings for system headers as well.

ClangSharp/Generated/CXCompletionContext.cs
mjsabby: Very nice!

mjsabby:
This is an exciting change! Thank you so much.

PR #47, opened 2019-05-12
Minor fixups to provide better diagnostics and notes on what should be done.

tannergooding:
This does some minor cleanup in the writer to ensure we are processing things like attributes correctly and adds some notes on where we could be doing better.

This also updates the main program to correctly print diagnostics and skip processing on error.

### DIFF
--- a/ClangSharpPInvokeGenerator/CursorWriter.cs
+++ b/ClangSharpPInvokeGenerator/CursorWriter.cs
@@ -141,9 +141,18 @@ namespace ClangSharpPInvokeGenerator
                 }
 
                 case CXCursorKind.CXCursor_UnexposedAttr:
+                {
+                    return BeginHandleUnexposedAttr(cursor, parent);
+                }
+
+                case CXCursorKind.CXCursor_DLLExport:
+                {
+                    return BeginHandleDLLExport(cursor, parent);
+                }
+
                 case CXCursorKind.CXCursor_DLLImport:
                 {
-                    return false;
+                    return BeginHandleDLLImport(cursor, parent);
                 }
 
                 default:
@@ -179,6 +188,7 @@ namespace ClangSharpPInvokeGenerator
                 case CXCursorKind.CXCursor_DeclRefExpr:
                 case CXCursorKind.CXCursor_IntegerLiteral:
                 case CXCursorKind.CXCursor_UnexposedAttr:
+                case CXCursorKind.CXCursor_DLLExport:
                 case CXCursorKind.CXCursor_DLLImport:
                 {
                     clearOutputBuilder = false;
@@ -373,6 +383,38 @@ namespace ClangSharpPInvokeGenerator
             }
         }
 
+        private bool BeginHandleDLLExport(CXCursor cursor, CXCursor parent)
+        {
+            Debug.Assert(cursor.Kind == CXCursorKind.CXCursor_DLLExport);
+
+            if (_attachedData.TryGetValue(parent, out var data) && (data is AttachedFunctionDeclData functionDeclData))
+            {
+                functionDeclData.IsDllExport = true;
+                return true;
+            }
+            else
+            {
+                Unhandled(cursor, parent);
+                return false;
+            }
+        }
+
+        private bool BeginHandleDLLImport(CXCursor cursor, CXCursor parent)
+        {
+            Debug.Assert(cursor.Kind == CXCursorKind.CXCursor_DLLImport);
+
+            if (_attachedData.TryGetValue(parent, out var data) && (data is AttachedFunctionDeclData functionDeclData))
+            {
+                functionDeclData.IsDllImport = true;
+                return true;
+            }
+            else
+            {
+                Unhandled(cursor, parent);
+                return false;
+            }
+        }
+
         private bool BeginHandleEnumConstantDecl(CXCursor cursor, CXCursor parent)
         {
             Debug.Assert(cursor.Kind == CXCursorKind.CXCursor_EnumConstantDecl);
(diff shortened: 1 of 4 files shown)


ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -100,20 +105,35 @@ public static int Run(InvocationContext context)
                 foreach (var file in files)
                 {
                     var translationUnitError = CXTranslationUnit.Parse(createIndex, file, arr, Array.Empty<CXUnsavedFile>(), translationFlags, out CXTranslationUnit translationUnit);
+                    bool skipProcessing = false;
 
                     if (translationUnitError != CXErrorCode.CXError_Success)
                     {
-                        Console.WriteLine($"Error: '{translationUnitError}' for '{file}'.");
-                        var numDiagnostics = translationUnit.NumDiagnostics;
+                        Console.WriteLine($"Error: Parsing failed for '{file}' due to '{translationUnitError}'.");
```
tannergooding: If the parsing failed, there is no guarantee the translation unit is even valid to get diagnostics

ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -100,20 +105,36 @@ public static int Run(InvocationContext context)
...
+                        skipProcessing = true;
+                    }
+                    else if (translationUnit.NumDiagnostics != 0)
+                    {
+                        Console.WriteLine($"Diagnostics for '{file}':");
```
tannergooding: A switch to not print warnings or to not skip on error could probably be provided in the future. For now though, clang and llvm generation both occur without any diagnostics and providing diagnostics for other files will help with resolving any codegen issues.

ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -100,20 +105,36 @@ public static int Run(InvocationContext context)
...
                         {
-                            var diagnostic = translationUnit.GetDiagnostic(i);
-                            Console.WriteLine(diagnostic.Spelling.ToString());
-                            diagnostic.Dispose();
+                            using (var diagnostic = translationUnit.GetDiagnostic(i))
+                            {
+                                Console.Write("    ");
+                                Console.WriteLine(diagnostic.Format(CXDiagnosticDisplayOptions.CXDiagnostic_DisplayOption).ToString());
+
+                                skipProcessing |= (diagnostic.Severity == CXDiagnosticSeverity.CXDiagnostic_Error);
```
tannergooding: Skipping processing on error/fatal diagnostics seems like the right default. Otherwise, you can end up with weird codegen due to missing types or clang misinterpreting a macro expansion as something else.

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -373,6 +383,38 @@ private bool BeginHandleDeclRefExpr(CXCursor cursor, CXCursor parent)
             }
         }
 
+        private bool BeginHandleDLLExport(CXCursor cursor, CXCursor parent)
+        {
+            Debug.Assert(cursor.Kind == CXCursorKind.CXCursor_DLLExport);
+
+            if (_attachedData.TryGetValue(parent, out var data) && (data is AttachedFunctionDeclData functionDeclData))
```
tannergooding: A dll export probably means you are processing the header with the wrong define. It might be useful to expose a way to surface diagnostics from our own processing up to the user...

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -373,6 +383,38 @@ private bool BeginHandleDeclRefExpr(CXCursor cursor, CXCursor parent)
...
+            }
+        }
+
+        private bool BeginHandleDLLImport(CXCursor cursor, CXCursor parent)
+        {
+            Debug.Assert(cursor.Kind == CXCursorKind.CXCursor_DLLImport);
+
+            if (_attachedData.TryGetValue(parent, out var data) && (data is AttachedFunctionDeclData functionDeclData))
+            {
+                functionDeclData.IsDllImport = true;
```
tannergooding: We should probably look for this as a requirement to emitting the method in the generated code. If it isn't there, then it might have a method body (for example) and it won't resolve at runtime.

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -932,6 +981,47 @@ private bool BeginHandleUnaryOperator(CXCursor cursor, CXCursor parent)
...
+
+                var identifier = tokens[0].GetSpelling(translationUnit).ToString();
+
+                if (identifier != "deprecated")
+                {
+                    Unhandled(cursor, parent);
+                    return false;
+                }
+
+                functionDeclData.IsDeprecated = true;
```
tannergooding: ClangSharp currently has just one of these and it would be nice to surface correctly. Maybe via the `[Obsolete]` attribute.

ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -83,13 +83,18 @@ public static int Run(InvocationContext context)
                 return -1;
             }
 
-            string[] arr = { "-x", "c++" };
+            var arr = new string[]
+            {
+                "-xc++",                                // The input files are C++
+                "-Wno-pragma-once-outside-header"       // We are processing files which may be header files
```
tannergooding: This is the one warning I think is common enough to always suppress.

Windows header files in general (e.g those from the Windows SDK) have a few other common warnings that it might be nice to provide a switch for easy suppression (such as #include directives using non matching casing).

PR #49, opened 2019-05-24
Implementing a structured traverser so the writer can have the full state

tannergooding:
This splits apart the visitor and the writer into two separate passes. Ultimately, this means that when writing files out to disk, we have the full state of the structure available for use.

During the initial traversal which builds the cursor tree, we are able to do some introspection about the cursor, such as being able to track the lhs and rhs of a binary operator, or being able to explicitly track function parameters or field declarations separately from other child cursors.

The writer then can worry about managing less state and only needs to track the output builders themselves plus which cursors have been visited.

This should also make it easier to process more complex expressions or statements in the writer.


tannergooding:
The cursor tree support is a bit verbose as I ensured it could build full trees for all the Windows SDK files. 

Most of the information isn't used anywhere yet as I preserved the existing behavior/handling that the writer had (so no changes if you regenerate ClangSharp or LLVMSharp).

tannergooding:
Another way of looking at this is it is taking the `Attached*Data` types that were being used and is filling them out to be more general purpose and carry all the information they can.

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -7,26 +7,19 @@
 
 namespace ClangSharpPInvokeGenerator
 {
-    internal class CursorWriter : CursorVisitor, IDisposable
+    internal sealed class CursorWriter : IDisposable
     {
-        private readonly Dictionary<CXCursor, object> _attachedData;
         private readonly Dictionary<string, OutputBuilder> _outputBuilders;
-        private readonly Stack<CXCursor> _processingCursors;
```
tannergooding: We don't need to track most of this state anymore.

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -41,1727 +34,1262 @@ public void Dispose()
...
-
-            _processingCursors.Push(cursor);
-            _visitedCursors.Add(cursor);
+            Debug.Assert(_outputBuilder is null);
+            _visitedCursors.Add(translationUnit);
+            VisitChildren(translationUnit);
+        }
 
-            switch (cursor.Kind)
+        private string EscapeName(string name)
```
tannergooding: The helper functions for generating names are the same as before, but pulling from the `Cursor` types rather than directly from `CXCursor`.

ClangSharpPInvokeGenerator/CursorWriter.cs
```diff
@@ -41,1727 +34,1262 @@ public void Dispose()
...
+                            }
+                            return name;
                         }
 
-                        case "uint16_t":
+                        default:
                         {
-                            return "ushort";
+                            Unhandled(decl, pointeeType);
+
```
tannergooding: Likewise for `ParenExpr`, plus we don't need to track how deep we are in the expression tree anymore.

tannergooding: This handles nested declarations, if they exist and places them just after the list of fields.

ClangSharpPInvokeGenerator/Cursors/Decls/StructDecl.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+
+                case CXCursorKind.CXCursor_EnumDecl:
+                {
+                    return GetOrAddChild<EnumDecl>(childHandle).Visit(clientData);
+                }
+
+                case CXCursorKind.CXCursor_FieldDecl:
+                {
+                    var fieldDecl = GetOrAddChild<FieldDecl>(childHandle);
```
tannergooding: Most code paths in these types aren't consumed by generating LLVMSharp or ClangSharp, but they can be used for generating things like the Windows SDK.

Child cursors that represent useful information for a given cursor kind can be exposed directly to the writer, such as the `FieldDecl` here. The writer can then handle that data first before falling back to normal processing for any remaining children.

ClangSharpPInvokeGenerator/Cursors/Exprs/IntegerLiteral.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    internal sealed class IntegerLiteral : Expr
+    {
+        private readonly Lazy<string> _rawValue;
+
+        public IntegerLiteral(CXCursor handle, Cursor parent) : base(handle, parent)
+        {
+            Debug.Assert(handle.Kind == CXCursorKind.CXCursor_IntegerLiteral);
+
+            _rawValue = new Lazy<string>(() => {
```
tannergooding: We can also relegate cursor specific logic, like extracting the raw text for an `IntegerLiteral` to the type itself, since it isn't really related to the writer.

ClangSharpPInvokeGenerator/Cursors/Exprs/UnaryOperator.cs
```diff
@@ -0,0 +1,190 @@
...
+        public Expr Expr
+        {
+            get
+            {
+                return _expr;
+            }
+
+            set
+            {
+                Debug.Assert(_expr is null);
```
tannergooding: We are also able to do additional sanity checks, such as an `UnaryOperator` only having a single expression (which can itself have sub-expressions).

mjsabby:
LGTM, took a while to review.

tannergooding:
> took a while to review.

Thanks, and sorry for the big PR. Subsequent ones should hopefully be smaller now and should mostly be restricted to the writer understanding more metadata from the traversed tree.

PR #51, opened 2019-05-26
Various improvements to make the code more useable for non-debug scenarios.

tannergooding:
This does a few things:
1. It splits ClangSharpPInvokeGenerator into the exe and a library that the exe calls into
2. It separates out most of the `OutputBuilder` creation logic into a factory class
3. It does a similar separation for the creation of the streams the output is written too, which allows use of in memory streams for the test scenario
4. It adds basic support for reporting diagnostics found when generating bindings, which allows users to see when the generator either doesn't support a cursor type or when it makes a possibly interesting decision.


ClangSharp.PInvokeGenerator.Test/PInvokeGeneratorTest.cs
```diff
@@ -0,0 +1,46 @@
...
+        protected static readonly string[] DefaultClangCommandLineArgs = new string[]
+        {
+            "-xc++",                                // The input files are C++
+            "-Wno-pragma-once-outside-header"       // We are processing files which may be header files
+        };
+
+        protected async Task ValidateGeneratedBindings(string inputContents, string expectedOutputContents)
+        {
+            // Ideally we would also have the input file be a MemoryStream. Unfortunately,
+            // Clang currently requires that all files exist on disk before Parse is called.
```
tannergooding: Everything is setup to enable this if Clang also supported it, but it turns out `CXUnsavedFile` requires `CXUnsavedFile.Filename` to already exit on disk.

Perhaps we could "make it work" by having an empty file with a well-known name as part of the build output for the test library...

ClangSharp.PInvokeGenerator/Cursors/Attrs/Attr.cs
```diff
@@ -1,7 +1,6 @@
 ﻿using System.Diagnostics;
-using ClangSharp;
 
-namespace ClangSharpPInvokeGenerator
+namespace ClangSharp
```
tannergooding: All of the structured traverser is still internal. It might be nice to make it public/reusable now that this is a separate library...

ClangSharp.PInvokeGenerator/Diagnostic.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace ClangSharp
+{
+    public sealed class Diagnostic
```
tannergooding: Diagnostics are really simple right now, but serve their purpose of reporting useful information well enough

ClangSharp.PInvokeGenerator/OutputBuilderFactory.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ClangSharp
+{
+    public sealed class OutputBuilderFactory
```
tannergooding: I thought about making this, `OutputBuilder`, and `PInvokeGenerator` abstract and then explicitly having a `CSharpOutputBuilderFactory`, `CSharpOutputBuilder`, and `CSharpPInvokeGenerator`. This would allow people to more trivially support other languages (like F#), if they so desired and would work fairly well since most of the logic is around the traversal, not the text generation...

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -3,43 +3,119 @@
...
-        private readonly ConfigurationOptions _config;
+        private readonly List<Diagnostic> _diagnostics;
+        private readonly PInvokeGeneratorConfiguration _config;
 
         private OutputBuilder _outputBuilder;
         private int _outputBuilderUsers;
+        private bool _disposed;
 
-        public CursorWriter(ConfigurationOptions config)
+        public PInvokeGenerator(PInvokeGeneratorConfiguration config, Func<string, (Stream stream, bool leaveOpen)> outputStreamFactory = null)
```
tannergooding: I'm not exactly pleased with the `outputStreamFactory`, but it seems to work. The alternative I came up with was an interface with a single method or something similar....

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -3,43 +3,119 @@
...
         }
 
-        public void Visit(TranslationUnit translationUnit)
+        public void Dispose()
+        {
+            Dispose(isDisposing: true);
+            GC.SuppressFinalize(this);
+        }
+
+        public void GenerateBindings(CXTranslationUnit translationUnitHandle)
```
tannergooding: This is now the primary entry point for the library and is what the executable calls into.

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -164,11 +345,31 @@ private string GetCallingConventionName(Cursor cursor, CXCallingConv callingConv
...
+                    return "Winapi";
+                }
+
                 default:
                 {
-                    Debug.WriteLine($"Unhandled calling convention: {callingConvention} in {cursor.KindSpelling}.");
-                    Debugger.Break();
-                    return string.Empty;
+                    var name = "WinApi";
+                    AddDiagnostic(DiagnosticLevel.Info, $"Unsupported calling convention: '{callingConvention}'. Falling back to '{name}'.", cursor);
```
tannergooding: All of the `Debugger.Break`/`Unhandle()` cases were replaced with diagnostics instead.

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -190,6 +391,7 @@ private string GetCursorName(Decl decl)
                             decl.Location.GetFileLocation(out var file, out var _, out var _, out var offset);
                             var fileName = Path.GetFileNameWithoutExtension(file.Name.ToString());
                             name = $"__Anonymous{decl.Type.KindSpelling}_{fileName}_{offset}";
+                            AddDiagnostic(DiagnosticLevel.Info, $"Anonymous declaration found in '{nameof(GetCursorName)}'. Falling back to '{name}'.'", decl);
```
tannergooding: Things like anonymous declarations or unsupported calling conventions are just `info`, since there isn't much we can do for it. We can really only generate code to the best of our ability.

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -301,8 +503,9 @@ private string GetCursorName(Decl decl)
 
                 default:
                 {
-                    Unhandled(decl);
-                    return string.Empty;
+                    var name = decl.Spelling;
+                    AddDiagnostic(DiagnosticLevel.Warning, $"Unsupported declaration: '{decl.KindSpelling}'. Falling back to '{name}'.", decl);
```
tannergooding: Things where the binding generation support is incomplete, but where we have a "reasonable" fallback are currently warnings. They generally represent cases where we might generate the correct code but it should really be investigated/handled properly.

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -972,16 +1162,10 @@ private void Visit(Cursor cursor, Cursor parent)
 
                 default:
                 {
-                    Debug.WriteLine($"Unhandled cursor kind: {cursor.KindSpelling} in {parent.KindSpelling}.");
-                    Debugger.Break();
+                    AddDiagnostic(DiagnosticLevel.Error, $"Unsupported cursor: '{cursor.KindSpelling}'. Generated bindings may be incomplete.", cursor);
```
tannergooding: Places where we can't make any reasonable fallback and we will actually not end up parsing the whole tree are surfaced as errors.

ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public sealed class PInvokeGeneratorConfiguration
+    {
+        private const string DefaultMethodClassName = "Methods";
+
+        private readonly PInvokeGeneratorConfigurationOptions _options;
+
+        public PInvokeGeneratorConfiguration(string libraryPath, string namespaceName, string outputLocation, PInvokeGeneratorConfigurationOptions options = PInvokeGeneratorConfigurationOptions.None, string[] excludedFunctions = null, string methodClassName = null, string methodPrefixToStrip = null)
+        {
+            if (excludedFunctions is null)
```
tannergooding: APIs that are now public are doing the appropriate input checking...

ClangSharp.PInvokeGenerator.Test/PInvokeGeneratorTest.cs
tannergooding: This did work, and I've pushed a commit so that everything is now "in memory".

tannergooding:
Added some more tests and validated that the generated code is consistent with what it was previously.

PR #52, opened 2019-05-26
Nuget metadata properties added to ClangSharp

manuzin:


ClangSharp/ClangSharp.csproj
```diff
@@ -3,6 +3,19 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <Description>ClangSharp are strongly-typed safe Clang bindings written in C# for .NET and Mono, tested on Linux and Windows. ClangSharp is self-hosted and ClangSharp auto-generates itself parsing LLVM-C header files.</Description>
+    <Version>3.9.0</Version>
```
tannergooding: @mjsabby, is this meant to be the same version as the `libclang` supported or is it independent?

ClangSharp/ClangSharp.csproj
```diff
@@ -3,6 +3,19 @@
...
+    <PackageLicenseUrl>https://github.com/microsoft/ClangSharp/blob/master/LICENSE.txt</PackageLicenseUrl>
+    <RepositoryUrl>https://github.com/microsoft/ClangSharp.git</RepositoryUrl>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageProjectUrl>http://www.clangsharp.org/</PackageProjectUrl>
   </PropertyGroup>
 
+  <ItemGroup>
+      <Content Include="ClangSharp.targets" Pack="true" PackagePath="build"/>
+      <Content Include="x64\libclang.dll" Pack="true" PackagePath="content\x64"/>
```
tannergooding: This isn't going to work correctly for linux...

tannergooding: In general, checking in binary files to git repos is considered "bad practice" and you should look at doing something else instead (such as pulling down a nuget package or using git lfs).

This should probably be pulling down the existing `libclang` package (but an updated version containing the 8.0 binaries).

manuzin: You are right, please ignore my changes, I will try to reuse libclang.runtime public package (I miss it).
The version (3.9.0) not related to libclang, current ClangSharp package available on nuget has 3.8.0 version so I simple use next version.

tannergooding: I think just using https://www.nuget.org/packages/Libclang/, which is published alongside ClangSharp by @mjsabby would be better.

That likely requires a new version to be published first, however.

manuzin: libclang uses 5.0 version, IMO it is too old, no?

tannergooding: Right, the current sources are based on `libclang 8.0.0`. So, it likely needs a new version to be published first.

mjsabby is the publisher of that package, and also the maintainer for this repo, so he should be able to do that 😄

manuzin: so lets wait :)

ClangSharp/ClangSharp.csproj
```diff
@@ -13,9 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-      <Content Include="ClangSharp.targets" Pack="true" PackagePath="build"/>
-      <Content Include="x64\libclang.dll" Pack="true" PackagePath="content\x64"/>
-      <Content Include="x86\libclang.dll" Pack="true" PackagePath="content\x86"/>
+    <PackageReference Include="libclang.runtime" Version="8.*" />
```
tannergooding: `*` versions are generally considered bad practice as it leads to non-deterministic builds. It also requires nuget to always do a server query as the version may have changed.

tannergooding:
Going to close this.

I ended up doing a large infrastructure rewrite to use Directory.Build.props/targets and most of this came naturally.

I then got the new `libclang v8.0.0` package published and had the dependency updated as part of doing that and getting CI up.

PR #53, opened 2019-05-27
Simplify the structured traverser and make it more closely mirror the C++ clang type structure

tannergooding:
This simplifies the structured traverser to not explicitly check each child type. Instead, each type will only special case the types where it is trying to pull relevant metadata.

This also updates the `Decl` types to more closely mirror the C++ clang type structure. For example, `StructDecl`, `ClassDecl`, and `UnionDecl` are all `CXXRecordDecl` types. This will allow more logic to be shared where it is relevant and makes it easier to determine which things expose certain properties.


ClangSharp.PInvokeGenerator/Cursors/Decls/CXXMethodDecl.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Diagnostics;
+
+namespace ClangSharp
+{
+    internal class CXXMethodDecl : FunctionDecl
```
tannergooding: For example, this follows the type hierarchy described here: https://clang.llvm.org/doxygen/classclang_1_1CXXMethodDecl.html.

This is: `Decl -> NamedDecl -> ValueDecl -> DeclaratorDecl -> FunctionDecl -> CXXMethodDecl`

tannergooding: This also means that we can centralize the `ParmDecls` property in `FunctionDecl`, rather then repeating it in `CXXMethod`, `Constructor`, `Destructor`, and `ConversionFunction`

ClangSharp.PInvokeGenerator/Cursors/Decls/ClassDecl.cs
```diff
@@ -2,124 +2,11 @@
 
 namespace ClangSharp
 {
-    internal sealed class ClassDecl : Decl
+    internal sealed class ClassDecl : CXXRecordDecl
```
tannergooding: A `ClassDecl` doesn't actually exist in the C++ type hierarchy, they are all just `RecordDecl` or `CXXRecordDecl`. I haven't finished investigating if that distinction is important for binding generation.

ClangSharp.PInvokeGenerator/Cursors/Decls/Decl.cs
```diff
@@ -158,16 +158,33 @@ public static new Decl Create(CXCursor handle, Cursor parent)
...
+
+            _definition = new Lazy<Cursor>(() => {
+                var cursor = TranslationUnit.GetOrCreateCursor(Handle.Definition, () => Create(Handle.Definition, this));
+                cursor.Visit(clientData: default);
+                return cursor;
+            });
         }
 
-        public bool IsAnonymous => Handle.IsAnonymous;
+        public CX_CXXAccessSpecifier AccessSpecifier => Handle.CXXAccessSpecifier;
```
tannergooding: I've started, but not completed exposing some of the `CXCursor` properties where they are actually relevant. For example, `CXCursor.CXXAccessSpecifier` only returns useful information for declarations and `CXXBaseSpecifier`, everything else just returns `CX_CXXInvalidAccessSpecifier`

ClangSharp.PInvokeGenerator/Cursors/Exprs/BinaryOperator.cs
```diff
@@ -59,118 +59,23 @@ protected override CXChildVisitResult VisitChildren(CXCursor childHandle, CXCurs
         {
             ValidateVisit(ref handle);
 
-            Expr expr;
-
-            switch (childHandle.Kind)
+            if (childHandle.IsExpression)
```
tannergooding: The logic is much more simplified now, and supports more. We know that a `BinaryOperator` will contain two children and they will both be `expressions`. The type of those expressions could be just about anything, so we really just want to check if it is one rather than check each type here.

ClangSharp.PInvokeGenerator/Cursors/Exprs/Expr.cs
```diff
@@ -138,14 +138,25 @@ public static new Expr Create(CXCursor handle, Cursor parent)
             }
         }
 
+        private readonly Lazy<Cursor> _definition;
         private readonly Lazy<Type> _type;
 
         protected Expr(CXCursor handle, Cursor parent) : base(handle, parent)
```
tannergooding: In the C++ type hierarchy, `Expr` is actually a sub-type of `Stmt`. I'll be looking at how to expose that later.

ClangSharp.PInvokeGenerator/Cursors/Decls/ClassDecl.cs
tannergooding: Doesn't look to be important outside of whether it is a Union or Class/Struct.

I've changed it to detect the language type to pick between `RecordDecl` and `CXXRecordDecl` and to expose `IsStruct`, `IsClass`, and `IsUnion` properties.

ClangSharp.PInvokeGenerator/Cursors/Decls/Decl.cs
tannergooding: I think I've completed exposing these properties for declarations now.

PR #54, opened 2019-05-29
Adding some tests validating we skip tag declarations that aren't definitions

tannergooding:
This fixes up the binding generation and adds corresponding tests validating that we will skip over any tag declarations which aren't actual definitions, but that we still emit any binding when no definition can be resolved.

This ensures that we still get a usable struct when we have opaque handles being defined, but also ensures we don't get any empty definition when that binding can be generated properly (either later as part of the same pass or by generating bindings over a separate header file).

### DIFF
--- a/ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
+++ b/ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
@@ -91,6 +91,58 @@ namespace ClangSharp.Test
             await ValidateGeneratedBindings(inputContents, expectedOutputContents);
         }
 
+        [Fact]
+        public async Task SkipNoDefinitionTest()
+        {
+            var inputContents = "typedef struct MyStruct MyStruct;";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public partial struct MyStruct
+    {{
+    }}
+}}
+";
+            await ValidateGeneratedBindings(inputContents, expectedOutputContents);
+        }
+
+        [Theory]
+        [InlineData("unsigned char", "byte")]
+        [InlineData("double", "double")]
+        [InlineData("short", "short")]
+        [InlineData("int", "int")]
+        [InlineData("long long", "long")]
+        [InlineData("char", "sbyte")]
+        [InlineData("float", "float")]
+        [InlineData("unsigned short", "ushort")]
+        [InlineData("unsigned int", "uint")]
+        [InlineData("unsigned long long", "ulong")]
+        public async Task SkipNonDefinitionTest(string nativeType, string expectedManagedType)
+        {
+            var inputContents = $@"typedef struct MyStruct MyStruct;
+
+struct MyStruct
+{{
+    {nativeType} r;
+    {nativeType} g;
+    {nativeType} b;
+}};
+";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public partial struct MyStruct
+    {{
+        public {expectedManagedType} r;
+        public {expectedManagedType} g;
+        public {expectedManagedType} b;
+    }}
+}}
+";
+
+            await ValidateGeneratedBindings(inputContents, expectedOutputContents);
+        }
+
         [Theory]
         [InlineData("unsigned char", "byte")]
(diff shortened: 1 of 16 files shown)


ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
```diff
@@ -91,6 +91,58 @@ public partial struct MyNestedStruct
             await ValidateGeneratedBindings(inputContents, expectedOutputContents);
         }
 
+        [Fact]
+        public async Task SkipNoDefinitionTest()
+        {
+            var inputContents = "typedef struct MyStruct MyStruct;";
```
tannergooding: This is the "opaque handle" case. Generally this will get consumed as a `MyStruct*` type elsewhere in the header file...

tannergooding: It might be useful to change the "excludeFunctions" option to be "excludeName" instead, which will exclude any named declaration with a matching name so users can skip over unnecessary structs/enums as well.

mjsabby:
Thanks for the explanation, makes sense.

ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
```diff
@@ -91,6 +91,58 @@ public partial struct MyNestedStruct
...
+        [InlineData("int", "int")]
+        [InlineData("long long", "long")]
+        [InlineData("char", "sbyte")]
+        [InlineData("float", "float")]
+        [InlineData("unsigned short", "ushort")]
+        [InlineData("unsigned int", "uint")]
+        [InlineData("unsigned long long", "ulong")]
+        public async Task SkipNonDefinitionTest(string nativeType, string expectedManagedType)
+        {
+            var inputContents = $@"typedef struct MyStruct MyStruct;
```
tannergooding: Here is a case where the definition exists later in the file. This ensures we only generate:
```cs
struct MyStruct
{
    float r;
}
```

rather than instead generating:
```cs
{
}

{
    float r;
    float g;
    float b;
}
```

ClangSharp.PInvokeGenerator/Cursors/Decls/Decl.cs
```diff
@@ -163,27 +163,21 @@ public static new Decl Create(CXCursor handle, Cursor parent)
...
-                return cursor;
+            _canonical = new Lazy<Decl>(() => {
+                var cursor = TranslationUnit.GetOrCreateCursor(Handle.CanonicalCursor, () => Create(Handle.CanonicalCursor, this));
+                cursor?.Visit(clientData: default);
+                return (Decl)cursor;
             });
             _lexicalParent = new Lazy<Cursor>(() => {
                 var cursor = TranslationUnit.GetOrCreateCursor(Handle.LexicalParent, () => Create(Handle.LexicalParent, this));
-                cursor.Visit(clientData: default);
+                cursor?.Visit(clientData: default);
```
tannergooding: Sometimes these are `null`, so we don't want to visit them in that scenario :)

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -79,7 +79,7 @@ public void Close()
                 emitNamespaceDeclaration = false;
             }
 
-            if (!_config.GenerateMultipleFiles)
+            if (!_config.GenerateMultipleFiles && _outputBuilderFactory.OutputBuilders.Any())
```
tannergooding: This avoids us from printing out just a `}` in the case of no `outputBuilders` being present

ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
mjsabby: Out of curiosity .. where did you find this type of declaration?

tannergooding: It happens fairly frequently in COM header files, like `dxgi.h`, as an example.

It is generally used to handle cases where you need to accept a method parameter before the type can actually be defined.

PR #55, opened 2019-05-29
Extending the exclude support and adding basic name remapping support

tannergooding:
This changes the `--excludeFunction` support to just be `--exclude` so that you can exclude types from binding generation as well.

This also adds a new `--remap name=value` feature that allows you to choose a different name to be emitted for types/cursors. This is particularly useful when the "true" name is less friendly than a more commonly used name (e.g. `tagRECT` vs `RECT` or `_MyStruct` vs `MyStruct`).


ClangSharpPInvokeGenerator/Properties/launchSettings.json
```diff
@@ -6,7 +6,7 @@
...
...e \"$(LLVMIncludePath)/llvm-c/TargetMachine.h\" --file \"$(LLVMIncludePath)/llvm-c/Types.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/AggressiveInstCombine.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/Coroutines.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/InstCombine.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/IPO.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/PassManagerBuilder.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/Scalar.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/Utils.h\" --file \"$(LLVMIncludePath)/llvm-c/Transforms/Vectorize.h\" -e LLVMDisasmInstruction -e LLVMGetBufferStart -e LLVMGetDiagInfoDescription -e LLVMGetDefaultTargetTriple -e LLVMCopyStringRepOfTargetData -e LLVMGetTargetMachineTriple -e LLVMGetTargetMachineCPU -e LLVMGetTargetMachineFeatureString -e LLVMPrintTypeToString -e LLVMCreateMessage -e LLVMDisposeMessage -e LLVMPrintModuleToString -e LLVMPrintValueToString -e LLVMOrcDisposeMangledSymbol -e LLVMTargetMachineEmitToFile -e LLVMInitializeCore -e LLVMAddInstructionCombiningPass -e LLVMAttributeIndex -e LLVMMetadataKind -r __AnonymousEnum_Core_16693=LLVMAttributeIndex -r __AnonymousEnum_DebugInfo_4207=LLVMMetadataKind"
```
tannergooding: The change here is `-e LLVMAttributeIndex -e LLVMMetadataKind -r __AnonymousEnum_Core_16693=LLVMAttributeIndex -r __AnonymousEnum_DebugInfo_4207=LLVMMetadataKind"` and means that anonymous declarations can be handled.

There is still likely some minor fixups if the header file changes significantly, but it is a step in the right direction.

ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -786,6 +792,30 @@ private string GetParmModifierForPointeeType(Decl decl, Type pointeeType)
             }
         }
 
+        private string GetRemappedCursorName(NamedDecl namedDecl)
```
tannergooding: There are both upsides and downsides to the super trivial approach here.

The upside is you can remap literally any name. The downside is that it impacts anything with that name and you may want a more precise filter (such as based on whether it is a "struct" or a "field").

PR #58, opened 2019-05-31
Adding some more improvements to the PInvokeGenerator

tannergooding:
This fixes the PInvokeGeneartor to handle single-file generation properly (namely, it was emitting using directives in the wrong place and overwriting the stream for each processed output builder).

This updates the PInvokeGenerator to emit a comment for variable declarations. Ideally this would be hooked up to the NativeLibraryLoader APIs, but that involves unsafe code and might need to be opt-in.

This also adds some tests for the fixed-sized buffer support and improves the codegen when the user has opted into unsafe code.

### DIFF
--- a/ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
+++ b/ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
@@ -62,6 +62,78 @@ namespace ClangSharp.Test
         [InlineData("unsigned short", "ushort")]
         [InlineData("unsigned int", "uint")]
         [InlineData("unsigned long long", "ulong")]
+        public async Task FixedSizedBufferNonPrimitiveTest(string nativeType, string expectedManagedType)
+        {
+            var inputContents = $@"struct MyStruct
+{{
+    {nativeType} value;
+}};
+
+struct MyOtherStruct
+{{
+    MyStruct c[3];
+}};
+";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public partial struct MyStruct
+    {{
+        public {expectedManagedType} value;
+    }}
+
+    public partial struct MyOtherStruct
+    {{
+        public MyStruct c0; public MyStruct c1; public MyStruct c2;
+    }}
+}}
+";
+
+            await ValidateGeneratedBindings(inputContents, expectedOutputContents);
+        }
+
+        [Theory]
+        [InlineData("unsigned char", "byte")]
+        [InlineData("double", "double")]
+        [InlineData("short", "short")]
+        [InlineData("int", "int")]
+        [InlineData("long long", "long")]
+        [InlineData("char", "sbyte")]
+        [InlineData("float", "float")]
+        [InlineData("unsigned short", "ushort")]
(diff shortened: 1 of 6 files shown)


ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
```diff
@@ -51,6 +51,78 @@ public async Task ExcludeTest()
...
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public partial struct MyStruct
+    {{
+        public {expectedManagedType} value;
+    }}
+
+    public partial struct MyOtherStruct
+    {{
+        public MyStruct c0; public MyStruct c1; public MyStruct c2;
```
tannergooding: This covers the existing generation logic for fixed-sized buffers.

ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
```diff
@@ -218,6 +290,89 @@ public partial struct MyStruct
...
+{{
+    {nativeType} c[3];
+}};
+";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public unsafe partial struct MyStruct
+    {{
+        public fixed {expectedManagedType} c[3];
```
tannergooding: This is the "ideal" generation and is what is used when the type is one of the types C# supports for fixed-sized buffers.

ClangSharp.PInvokeGenerator.Test/StructDeclarationTest.cs
```diff
@@ -218,6 +290,89 @@ public partial struct MyStruct
...
+    {{
+        public _c_e__FixedBuffer c;
+
+        public partial struct _c_e__FixedBuffer
+        {{
+            private MyStruct e0;
+            private MyStruct e1;
+            private MyStruct e2;
+
+            public ref MyStruct this[int index] => ref MemoryMarshal.CreateSpan(ref e0, 3)[index];
```
tannergooding: This is the fallback generation (still for unsafe code) for when C# doesn't support the type for "proper" fixed-sized buffers.

It takes advantage of `Span` and `ref returns` to avoid pinning and allow "normal" use of the API.

ClangSharp.PInvokeGenerator.Test/VarDeclarationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var inputContents = $@"{nativeType} MyVariable;";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public static partial class Methods
+    {{
+        private const string libraryPath = ""ClangSharpPInvokeGenerator"";
+
+        // public static extern {expectedManagedType} MyVariable;
```
tannergooding: As mentioned, ideally this would be an `IntPtr` or a `T*` (depending on if safe or unsafe code) and would use the native library loader APIs (only available in .NET Core 3 and higher) to initialize the handle/address correctly.

I'll likely fix this up in a future PR, but for now, it lets consumers more easily see when these variables exist and take action themselves.

PR #60, opened 2019-05-31
Setup some infrastructure for the repository.

tannergooding:
This sets up a lot of infrastructure for the repository including using Directory.Build.props/targets and adding various scripts for building, restoring, testing, and packing.

I believe the ultimate end-goal is to get on dotnet/arcade, but this should still be a step in the right direction.

It does not currently enable strong-name signing (as it needs to be determined if a custom or existing strong name key will be used). It would also be trivial to hook up authenticode signing once that is resolved.

If we have an existing AzDO repo, I can also add the `azure-pipelines.yml` file and get that enabled.


.editorconfig
```diff
@@ -1,6 +1,371 @@
...
+#    suggest const declarations be pascal case
+#    suggest interfaces be pascal case prefixed with I
+#    suggest parameters be camel case
+#    suggest private and internal static fields be camel case
+#    suggest private and internal fields be camel case
+#    suggest public and protected declarations be pascal case
+#    suggest static readonly declarations be pascal case
+#    suggest type parameters be prefixed with T
+###############################################################################
+[*.cs]
```
tannergooding: These all match the VS defaults modulo `csharp_indent_case_contents_when_block`

.editorconfig
```diff
@@ -1,6 +1,371 @@
...
+#   indent case contents,
+#   not indent case contents when block,
+#   indent labels one less than the current, and
+#   indent switch labels
+###############################################################################
+[*.cs]
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = false
```
tannergooding: This is the only new option that differs from the VS default. The VS default is `true` for compatibility reasons. This option is used in combination with `csharp_indent_case_contents` and you have effectively the following combinations:

(VS Default) `csharp_indent_case_contents=true`, `csharp_indent_case_contents_when_block=true`:
```cs
switch (value)
{
    case 1:
        break;

    case 2:
        {
            break;
        }
}
```

`csharp_indent_case_contents=false`, `csharp_indent_case_contents_when_block=true`:
```cs
switch (value)
{
    case 1:
    break;

    case 2:
        {
            break;
        }
}
```

`csharp_indent_case_contents=false`, `csharp_indent_case_contents_when_block=false`:
```cs
switch (value)
{
    case 1:
    break;

    case 2:
    {
        break;
    }
}
```

(The one I chose) `csharp_indent_case_contents=true`, `csharp_indent_case_contents_when_block=false`:
```cs
switch (value)
{
    case 1:
        break;

    case 2:
    {
        break;
    }
}
```

.gitignore
```diff
@@ -1,186 +1,16 @@
-## Ignore Visual Studio temporary files, build results, and
-## files generated by popular Visual Studio add-ons.
-
-# User-specific files
-*.suo
-*.user
-*.sln.docstates
+###############################################################################
```
tannergooding: The main `.gitignore` file has a lot more than necessary. I can keep it if desired, but I personally find the smaller subset easier to reason about.

tannergooding: (It also helps expose places where things aren't ending up where expected as almost all excluded files should end up under `artifacts`).

Directory.Build.props
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
```
tannergooding: Centralized basically all of the relevant project settings here.

Directory.Build.props
```diff
@@ -0,0 +1,73 @@
...
+  </PropertyGroup>
+
+  <!-- Default settings that explicitly differ from the Sdk.targets defaults-->
+  <PropertyGroup>
+    <Authors>Microsoft and Contributors</Authors>
+    <BaseOutputPath>$(BaseArtifactsPath)bin/$(BaseArtifactsPathSuffix)/</BaseOutputPath>
+    <Company>Microsoft</Company>
+    <PackageOutputPath>$(BaseArtifactsPath)pkg/$(BaseArtifactsPathSuffix)/$(Configuration)/</PackageOutputPath>
+    <Product>ClangSharp</Product>
+    <VersionPrefix>4.0.0</VersionPrefix>
```
tannergooding: Not sure what version is desired. The last NuGet package was 3.8.0

Directory.Build.props
```diff
@@ -0,0 +1,73 @@
...
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+    <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
+  </PropertyGroup>
+
+  <!-- Default settings that explicitly differ from the Sdk.targets defaults-->
+  <PropertyGroup>
+    <Authors>Microsoft and Contributors</Authors>
+    <BaseOutputPath>$(BaseArtifactsPath)bin/$(BaseArtifactsPathSuffix)/</BaseOutputPath>
+    <Company>Microsoft</Company>
```
tannergooding: Should this be `Mukul Sabharwal` like in the license file or `Microsoft`, given it is in the `microsoft` github org?

Directory.Build.targets
```diff
@@ -0,0 +1,32 @@
...
+    <DefineConstants>$(DefineConstants);$(OS)</DefineConstants>
+  </PropertyGroup>
+
+  <!-- Tool versions for tool references across all projects -->
+  <ItemGroup>
+    <DotNetCliToolReference Update="dotnet-xunit" Version="2.3.1" />
+  </ItemGroup>
+
+  <!-- Package versions for package references across all projects -->
+  <ItemGroup>
```
tannergooding: Package versions are centrally managed here.

mjsabby:
What do you mean by existing AzDO repo? Do we need a repo? I thought we just need an instance.

tannergooding:
Right, I meant instance. Something like you have for the LLVM repo.

Directory.Build.props
mjsabby: Microsoft

mjsabby:
Ok ... I'd love if it could be in dnceng .

Directory.Build.props
mjsabby: Well won't this be 8.0? I was thinking we should keep the libClang versions

.gitignore
mjsabby: I agree ... I just let GitHub decide so this is better imo too.

.editorconfig
mjsabby: I personally don't like the corefx style, but since you've written most of the code here now you can pick whatever you want.

Directory.Build.props
tannergooding: We can certainly do that, will update.

tannergooding:
> Ok ... I'd love if it could be in dnceng 

I've pinged the e-mail thread to see what is recommended here.

Directory.Build.props
tannergooding: Fixed.

Directory.Build.props
tannergooding: Should the license file likewise be updated to be `Microsoft` and/or `Microsoft and Contributors`?

.editorconfig
tannergooding: Which one of the four is your preference? I don't feel particularly strongly as long as the editorconfig ensures it is automatically normalized.

Directory.Build.props
mjsabby: yup!

.editorconfig
mjsabby: The last one seems ok. I don't know why it's not the default. The VS Default looks ugly.

Directory.Build.props
tannergooding: Fixed.

.editorconfig
tannergooding: The last one is the one I already choose for the repo (and was the one I thought you were indicating you didn't like) 😄

The first is the VS default for compatibility reasons at this point. The option allowing the last setup was only introduced a couple versions ago (I think VS2015 or 2017) and hasn't been around for very long at this point.

PR #61, opened 2019-06-05
Cleanup how the PInvokeGenerator handles CXType and resolves names

tannergooding:
This does some minor cleanup that namely extends the CXType wrapper to more closely match the Clang C++ bindings.

This also improves how cursor and type names are resolved to better match how the wrapper types/etc are generated.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -80,6 +80,8 @@ namespace ClangSharp
                 {
                     using (var sw = new StreamWriter(stream, DefaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen))
                     {
+                        sw.NewLine = "\n";
+
                         foreach (var usingDirective in usingDirectives)
                         {
                             sw.Write("using");
@@ -123,6 +125,7 @@ namespace ClangSharp
 
                 using (var sw = new StreamWriter(stream, DefaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen))
                 {
+                    sw.NewLine = "\n";
                     sw.WriteLine('}');
                 }
             }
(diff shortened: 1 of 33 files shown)


tannergooding:
This also sets things up so that support for generating "safe code" can be removed. This support, while nice, is very error prone and can have various hidden overhead that isn't going to be easy to maintain long term.

Namely, things like `out` vs `ref` and marshaling behavior (especially around things like arrays) is almost impossible to determine without some kind of additional input. Wrapper types are also not always compatible with the underlying calling convention (namely for `thiscall`) and can prevent an API from working as expected.

The best case scenario is to just generate bindings which have 1-to-1 mapping with the underlying signatures and with no marshaling. This gives a zero-cost (modulo the overhead of the P/Invoke stub) abstraction over the native APIs which allows usage in performance oriented scenarios and allows better use with new library functionality such as `Span<T>`. Safe managed wrapper APIs can then be created over those unsafe native bindings (as we're already doing in ClangSharp/LLVMSharp) which expose the friendly API.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -843,299 +729,336 @@ private string GetParmModifierForPointeeType(Decl decl, Type pointeeType)
...
+                        break;
+                    }
 
-                            if (_config.GenerateUnsafeCode)
-                            {
-                                name += '*';
-                            }
-                            return name;
-                        }
+                    case "size_t":
```
tannergooding: `size_t` is spec'd to be unsigned, so this is a change in how the bindings are generated.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -843,299 +729,336 @@ private string GetParmModifierForPointeeType(Decl decl, Type pointeeType)
...
-                case CXTypeKind.CXType_Float:
-                case CXTypeKind.CXType_Double:
-                case CXTypeKind.CXType_Pointer:
-                case CXTypeKind.CXType_Record:
-                case CXTypeKind.CXType_Enum:
-                case CXTypeKind.CXType_Typedef:
-                {
-                    switch (decl.Kind)
+                    case "int64_t":
+                    case "time_t":
```
tannergooding: `time_t` is generally 64-bit on modern hardware (even 32-bit machines), but not always. Should probably provide a way to override this....

PR #62, opened 2019-06-06
Adding an azure-pipelines.yml file.

tannergooding:
Adding a basic `azure-pipelines.yml` file so that CI can be enabled.


tannergooding:
After this is done, I can add the build pipeline to the AzDO instance and submit another PR adding the status badges to the README/validating everything works.

PR #63, opened 2019-06-06
Adding build status badges to the README.md

tannergooding:
This also fixes up some issues that were preventing CI from running.

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -110,7 +110,9 @@ try {
   Create-Directory -Path $LogDir
 
   if ($architecture -ne "") {
-    $DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
+    $env:DOTNET_CLI_TELEMETRY_OPTOUT = 1
+    $env:DOTNET_MULTILEVEL_LOOKUP = 0
+    $env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
 
     $DotNetInstallScript = Join-Path -Path $ArtifactsDir -ChildPath "dotnet-install.ps1"
     Invoke-WebRequest -Uri "https://dot.net/v1/dotnet-install.ps1" -OutFile $DotNetInstallScript -UseBasicParsing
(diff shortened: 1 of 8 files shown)


tests/ClangSharp.PInvokeGenerator.UnitTests/EnumDeclarationTest.cs
```diff
@@ -98,31 +98,24 @@ public enum MyEnum : {expectedManagedType}
             await ValidateGeneratedBindings(inputContents, expectedOutputContents);
         }
 
-        [Fact]
-        public async Task NoDefinitionTest()
```
tannergooding: Clang 5 reports this as a "Microsoft Extension" and doesn't actually add a declaration for the enum so it results in no codegen.

tests/ClangSharp.PInvokeGenerator.UnitTests/EnumDeclarationTest.cs
```diff
@@ -9,7 +9,7 @@ public sealed class EnumDeclarationTest : PInvokeGeneratorTest
         [Fact]
         public async Task BasicTest()
         {
-            var inputContents = @"enum MyEnum
+            var inputContents = @"enum MyEnum : int
```
tannergooding: Unix defaults to `: uint`, marking it explicitly `int` so tests pass for now.

tests/ClangSharp.PInvokeGenerator.UnitTests/EnumDeclarationTest.cs
tannergooding: (specifically it marks forward declarations of enums as a "Microsoft Extension")

scripts/build.ps1
```diff
@@ -110,16 +110,19 @@ try {
   Create-Directory -Path $LogDir
 
   if ($architecture -ne "") {
-    $DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
+    $env:DOTNET_CLI_TELEMETRY_OPTOUT = 1
```
tannergooding: This wasn't be respected if it wasn't properly exported. Will also submit a fix for LLVMSharp.

sources/ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -100,6 +100,7 @@ public static int Run(InvocationContext context)
 
             var clangCommandLineArgs = new string[]
             {
+                "-std=c++11",                           // The input files should be compiled for C++ 11
```
tannergooding: Clang 5 (which is currently being used for tests) doesn't target C++11 by default yet.

PR #65, opened 2019-06-07
Fixing up some infrastructure so we can consume the libclang 8.0.0 packages

tannergooding:
This adds some nuspecs for the libclang package to make it easier to create in the future. This also does some misc infrastructure upgrades to make it easier to consume the package once published.


Directory.Build.props
```diff
@@ -67,7 +67,7 @@
 
   <!-- Package references which are consumed by all projects -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.Net.Compilers.Toolset" IsImplicitlyDefined="true" />
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" IsImplicitlyDefined="true" PrivateAssets="all" />
```
tannergooding: Needed to mark this as `PrivateAssets="all"` so that the product packages don't list it as a dependency.

Directory.Build.targets
```diff
@@ -15,6 +15,11 @@
     <DefineConstants>$(DefineConstants);$(OS)</DefineConstants>
   </PropertyGroup>
 
+  <!-- Settings that allow testing to work by default -->
+  <PropertyGroup>
+    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
```
tannergooding: The libclang package is setup using a `runtime.json` file, so tests can't run without a `RID` being specified (so the correct libclang` native dependency can be resolved). This sets the default to just be the same as the host SDK.

Directory.Build.targets
```diff
@@ -24,10 +29,10 @@
   <ItemGroup>
     <PackageReference Update="libClang" Version="5.0.0" NoWarn="NU1701" />
     <PackageReference Update="Microsoft.Net.Compilers.Toolset" Version="3.2.0-beta1-final" />
-    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="15.6.0" />
+    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.1.1" />
     <PackageReference Update="System.CommandLine.Experimental" Version="0.2.0-alpha.19174.3" />
-    <PackageReference Update="xunit" Version="2.3.1" />
-    <PackageReference Update="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Update="xunit" Version="2.4.1" />
```
tannergooding: These were a few versions old. Upgrading to get perf and other minor benefits.

packages/libclang/LICENSE.TXT
```diff
@@ -0,0 +1,278 @@
+==============================================================================
```
tannergooding: LICENSE.TXT was taken directly from https://github.com/llvm-mirror/clang

packages/libclang/libclang.nuspec
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata minClientVersion="2.12">
+    <id>libclang</id>
+    <version>8.0.0</version>
+    <authors>Microsoft and Contributors</authors>
```
tannergooding: NuGet lists this as being the package authors/owners, not the binary authors/owners.

packages/libclang/runtime.json
```diff
@@ -0,0 +1,59 @@
+{
+  "runtimes": {
```
tannergooding: LLVM supports 12 platforms, of those, .NET Core supports 11 (no `PowerPC` support).

packages/libclang/runtime.json
```diff
@@ -0,0 +1,59 @@
...
+      "libclang": {
+        "runtime.linux-arm64.libclang": "8.0.0"
+      }
+    },
+    "osx-x64": {
+      "libclang": {
+        "runtime.osx-x64.libclang": "8.0.0"
+      }
+    },
+    "sles-x64": {
```
tannergooding: The native dll is technically `sles.11.3`, but that isn't a valid RID

tannergooding: .NET supports `sles.12` as the earliest version

sources/ClangSharpPInvokeGenerator/ClangSharpPInvokeGenerator.csproj
```diff
@@ -1,9 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
tannergooding: netcoreapp2.0 went EOL (end of life) back in October of 2018

tannergooding:
Once this is approved/merged, I can publish the libclang.8.0.0 nuget packages and can submit the PR consuming them.

libllvm is going to be a bit more complex as it requires building per RID (which is feasible, it just will take a bit of time 😄).

packages/libclang/libclang.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <owners>Microsoft and Contributors</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="expression">Apache-2.0</license>
+    <projectUrl>https://github.com/llvm-mirror/clang</projectUrl>
```
tannergooding: The license should definitely match the original binaries, but it might make sense to use the clangsharp project url, instead of theirs. Thoughts?

tannergooding: There is also the repository url...

tannergooding: Actually, looks like NuGet lists `projectUrl` as
> A URL for the package's home page, often shown in UI displays as well as nuget.org.

`repository` is listed as:
> Repository metadata, consisting of four optional attributes: type and url (4.0+), and branch and commit (4.6+). These attributes allow you to map the .nupkg to the repository that built it, with the potential to get as detailed as the individual branch or commit that built the package. This should be a publicly available url that can be invoked directly by a version control software. It should not be an html page as this is meant for the computer. For linking to project page, use the projectUrl field, instead.

@mjsabby, what do you think of making `projectUrl` be `https://github.com/microsoft/ClangSharp` and `repository` be: `<repository type="git" url="https://github.com/llvm-mirror/clang" branch="release_80" />`?

I think that would satisfy both who created the package and where the sources were built from.

mjsabby: Sounds good.

tannergooding: Fixed that up, and also updated the nuspec to:
1. Explicitly list the content files (so a failure occurs if you forget to copy over the native library)
2. Reuse the same LICENSE.TXT file from the `packages/libclang` directory, so we don't need 12 copies of it.

PR #67, opened 2019-06-07
Changing `runtime.rid.libclang` to be `libclang.runtime.rid`

tannergooding:
The runtime prefix is reserved by nuget.

### DIFF
--- a/ClangSharp.sln
+++ b/ClangSharp.sln
@@ -52,70 +52,59 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "libclang", "libclang", "{C8
 		packages\libclang\runtime.json = packages\libclang\runtime.json
 	EndProjectSection
 EndProject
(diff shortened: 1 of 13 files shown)


Directory.Build.targets
```diff
@@ -27,7 +27,7 @@
 
   <!-- Package versions for package references across all projects -->
   <ItemGroup>
-    <PackageReference Update="libClang" Version="5.0.0" NoWarn="NU1701" />
+    <PackageReference Update="libclang" Version="8.0.0" />
```
tannergooding: Also updated this to use the new package.

tannergooding:
Will determine what dependency the ubuntu CI machines are missing later tonight.

Directory.Build.targets
tannergooding: Dropped this so I can put it in it's own PR and not block fixing the nuspecs.

PR #68, opened 2019-06-08
Updating libclang to use the new 8.0.0 package.

tannergooding:
This looks to require a fix to the ubuntu 16.04 images as they are missing a dependency.

### DIFF
--- a/scripts/azure-pipelines.yml
+++ b/scripts/azure-pipelines.yml
@@ -52,6 +52,8 @@ jobs:
 - job: ubuntu_1604_debug_x64
   pool:
     name: Hosted Ubuntu 1604
+  variables:
+    RuntimeIdentifier: ubuntu.16.04-x64
   steps:
   - task: Bash@3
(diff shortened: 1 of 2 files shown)


scripts/azure-pipelines.yml
```diff
@@ -52,6 +52,8 @@ jobs:
 - job: ubuntu_1604_debug_x64
   pool:
     name: Hosted Ubuntu 1604
+  variables:
+    RuntimeIdentifier: ubuntu.16.04-x64
```
tannergooding: The `Directory.Build.targets` is setting the default `RuntimeIdentifier` to `NETCoreSdkRuntimeIdentifier`. However, it appears that defaults to `linux-x64` and not the same RID as is returned from `dotnet --info` (which would be `ubuntu.16.04-x64` on these machines).

This will workaround the issue and allow CI to work until an appropriate fix can be resolved (and I've started a thread to try and determine what can be done here).

PR #69, opened 2019-06-11
Drop support for generating "safe" code

tannergooding:
As called out [here](https://github.com/microsoft/ClangSharp/pull/61#issuecomment-499188376), the current support for generating "safe" code is very error prone and isn't going to be easy to maintain long term.

Namely, almost all of the support is a "best" guess heuristic where the heuristics can vary wildly from header file to header file and we frequently get it wrong, especially when dealing with `bool`, `out`, `ref`, or `arrays`.

This removes that support and switches to only generating `unsafe` code. As part of that, it updates the logic of when the `unsafe` keyword is emitted to ensure it is only emitted when actually required. This also adds `NativeTypeName` attribute metadata to members where the native type and managed type do not line up (this could do with some tuning in the future, as it is currently overly conservative). The attribute is marked as `[Conditional("DEBUG")]` so it doesn't bloat the release binaries.

This then regenerates ClangSharp and fixes up the higher level bindings where required.


sources/ClangSharp.PInvokeGenerator/Cursors/Cursor.cs
```diff
@@ -82,10 +83,9 @@ protected Cursor(CXCursor handle, Cursor parent)
 
         public TranslationUnit TranslationUnit { get; }
 
-        public CXToken[] Tokenize(Cursor cursor)
+        public Span<CXToken> Tokenize(Cursor cursor)
```
tannergooding: Took a dependency on `System.Memory` to avoid creating arrays and copying overhead where possible.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -422,6 +463,54 @@ private string EscapeAndStripName(string name)
             return EscapeName(name);
         }
 
+        private string GetAccessSpecifierName(NamedDecl namedDecl)
```
tannergooding: Also added support for matching the native access specifier.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -747,19 +605,20 @@ private string GetRemappedName(string name)
...
-        private string GetTypeName(NamedDecl namedDecl, Type type)
+        private string GetTypeName(NamedDecl namedDecl, Type type, out string nativeTypeName)
         {
             var name = type.Spelling;
+            nativeTypeName = name;
 
             if (type is ArrayType arrayType)
             {
-                name = GetTypeName(namedDecl, arrayType.ElementType);
+                name = GetTypeName(namedDecl, arrayType.ElementType, out var nativeElementTypeName);
```
tannergooding: The secondary variables here are only used for debugging purposes. In some future PR, they might be used to drop more `NativeTypeNameAttributes` when possible (but the logic for treating `unsigned int *` as equal to `uint*` or transforming `const unsigned int * const *` to be `const uint * const *` is a bit tricky).

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -866,55 +725,31 @@ private string GetTypeName(NamedDecl namedDecl, Type type)
...
-                    }
-                }
+                // We intercept some well known types that have variable sizes
+                // to ensure that we can treat them correctly. Otherwise, they
+                // will resolve to a particular platform size, based on whatever
+                // parameters were passed into clang.
 
                 switch (name)
                 {
-                    case "int8_t":
```
tannergooding: All of these "just work" now.

sources/ClangSharp/Extensions/CXCompletionString.cs
```diff
@@ -14,12 +14,18 @@ public partial struct CXCompletionString
...
+        public CXCompletionString GetChunkCompletionString(uint index) => (CXCompletionString)clang.getCompletionChunkCompletionString(this, index);
 
         public CXCompletionChunkKind GetChunkKind(uint index) => clang.getCompletionChunkKind(this, index);
 
         public CXString GetChunkText(uint index) => clang.getCompletionChunkText(this, index);
 
-        public CXString GetParent(ref CXCursorKind kind) => clang.getCompletionParent(this, ref kind);
+        public CXString GetParent(out CXCursorKind kind)
+        {
+            fixed (CXCursorKind* pKind = &kind)
```
tannergooding: The downside (or upside, depending on how you look at it) to having unsafe bindings is you get to explicitly see all the "hidden" overhead that the marshaler was adding (such as implicitly pinning a value).

sources/ClangSharp/Extensions/CXModuleMapDescriptor.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+
+namespace ClangSharp
+{
+    public unsafe partial struct CXModuleMapDescriptor : IDisposable
```
tannergooding: Had to add a couple more "safe" wrappers so the non-PInvokeGenerator unit tests would pass.

sources/ClangSharp/Extensions/CXModuleMapDescriptor.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (Pointer != IntPtr.Zero)
+            {
+                clang.ModuleMapDescriptor_dispose(this);
+                Pointer = IntPtr.Zero;
+            }
+        }
+
+        public CXErrorCode SetFrameworkModuleName(string name)
+        {
+            using (var marshaledName = new MarshaledString(name))
```
tannergooding: This is how I opted to do the string marshaling (and cleanup) currently.

Ideally the "safe" API would take a `UTF8String` rather than `string`, but that type doesn't exist yet (and `ReadOnlySpan<byte>` support is fairly limited in `netstandard2.0`).

sources/ClangSharp/Extensions/CXString.cs
```diff
@@ -1,10 +1,24 @@
...
+            {
+                var pCString = clang.getCString(this);
+
+                if (pCString is null)
+                {
+                    return string.Empty;
+                }
+
+                var span = new ReadOnlySpan<byte>(pCString, int.MaxValue);
+                return span.Slice(0, span.IndexOf((byte)'\0')).AsString();
```
tannergooding: Marshaling from native to `string` just utilizes `ROSpan<byte>` and `System.Text.Encoding.UTF8.GetBytes`

sources/ClangSharp/Extensions/CXTranslationUnit.cs
```diff
@@ -1,28 +1,84 @@
...
+            }
+        }
+
+        public static CXTranslationUnit CreateFromSourceFile(CXIndex index, string sourceFileName, string[] commandLineArgs, CXUnsavedFile[] unsavedFiles)
+        {
+            using (var marshaledSourceFileName = new MarshaledString(sourceFileName))
+            using (var marshaledCommandLineArgs = new MarshaledStringArray(commandLineArgs))
+            fixed (CXUnsavedFile* pUnsavedFiles = unsavedFiles)
+            {
+                var pCommandLineArgs = stackalloc sbyte*[marshaledCommandLineArgs.Count];
```
tannergooding: Marshaling string arrays is currently a bit more complex. I had tried to allocate a `sbyte**` and then fill it directly (to avoid needing to allocate a `sbyte*[]`), but that was having problems and I decided to resolve it later.

sources/ClangSharp/Extensions/CXTranslationUnit.cs
```diff
@@ -32,45 +88,100 @@ public partial struct CXTranslationUnit : IDisposable
 
         public CXTargetInfo TargetInfo => clang.getTranslationUnitTargetInfo(this);
 
-        public void AnnotateTokens(CXToken[] tokens, CXCursor[] cursors) => clang.annotateTokens(this, tokens, (uint)tokens.Length, cursors);
-
-        public void Dispose() => clang.disposeTranslationUnit(this);
-
-        public void DisposeTokens(CXToken[] tokens) => clang.disposeTokens(this, tokens, (uint)tokens.Length);
+        public void AnnotateTokens(CXToken[] tokens, CXCursor[] cursors)
```
tannergooding: It would be nice if most of these were updated to take a `ROSpan<T>` or `Span<T>`, as appropriate. So you can avoid allocations where appropriate, but I left that to a future PR.

sources/ClangSharp/Generated.Custom.cs
```diff
@@ -1,325 +0,0 @@
-namespace ClangSharp
```
tannergooding: I was able to get rid of all method/type exclusions and just generate the bindings directly instead.

sources/ClangSharp/Generated/CXCompletionResult.cs
```diff
@@ -1,8 +1,11 @@
 namespace ClangSharp
 {
-    public partial struct CXCompletionResult
+    public unsafe partial struct CXCompletionResult
     {
+        [NativeTypeName("enum CXCursorKind")]
```
tannergooding: Another example of where "normalizing" the native type name would allow us to drop the attribute.

sources/ClangSharp/Generated/CXCursor.cs
```diff
@@ -1,11 +1,31 @@
...
     public partial struct CXCursor
     {
+        [NativeTypeName("enum CXCursorKind")]
         public CXCursorKind kind;
+
         public int xdata;
-        public IntPtr data0; public IntPtr data1; public IntPtr data2;
+
+        [NativeTypeName("const void *[3]")]
+        public _data_e__FixedBuffer data;
```
tannergooding: I find the attributes to, at a glance, make it much easier to determine the original signature. While still allowing you to use the least expensive managed signature. Here is a good example.

sources/ClangSharp/Generated/CXIdxObjCInterfaceDeclInfo.cs
```diff
@@ -1,11 +1,14 @@
...
+    public unsafe partial struct CXIdxObjCInterfaceDeclInfo
     {
-        public IntPtr containerInfo;
-        public IntPtr superInfo;
-        public IntPtr protocols;
+        [NativeTypeName("const CXIdxObjCContainerDeclInfo *")]
+        public CXIdxObjCContainerDeclInfo* containerInfo;
+
+        [NativeTypeName("const CXIdxBaseClassInfo *")]
+        public CXIdxBaseClassInfo* superInfo;
```
tannergooding: It also allows you to see the "const" info, which can be very useful for exposing the right managed API (and knowing things like: "should this be span or rospan").

sources/ClangSharp/Internals/NativeTypeNameAttribute.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Diagnostics;
+
+namespace ClangSharp
+{
+    [Conditional("DEBUG")]
```
tannergooding: `Conditional("DEBUG")` will cause the attribute to be stripped from release binaries, avoiding bloat.

sources/ClangSharp/Wrappers/CXCursorSet.cs
```diff
@@ -0,0 +1,24 @@
...
+    public unsafe partial struct CXCursorSet
+    {
+        public CXCursorSet(IntPtr pointer)
+        {
+            Pointer = pointer;
+        }
+
+        public IntPtr Pointer;
+
+        public static implicit operator CXCursorSet(CXCursorSetImpl* value)
```
tannergooding: Using implicit operators to convert from the native type to the wrapper type when it is an "exact" match.

sources/ClangSharp/Wrappers/CXDiagnostic.cs
```diff
@@ -0,0 +1,24 @@
...
+    public unsafe partial struct CXDiagnostic
+    {
+        public CXDiagnostic(IntPtr pointer)
+        {
+            Pointer = pointer;
+        }
+
+        public IntPtr Pointer;
+
+        public static explicit operator CXDiagnostic(void* value)
```
tannergooding: Using explicit operators to convert from the native type to the wrapper type when it is an "ambiguous" match.

PR #72, opened 2019-06-14
 Ensure APIs are exposed in the simple interop extensions.

tannergooding:
This resolves https://github.com/microsoft/ClangSharp/issues/33.

This does a few steps:
1. It moves the generated bindings into the ClangSharp.Interop namespace
2. It merges the wrappers and extensions folders into a `Interop.Extensions` folder
3. It goes through the `clang` class and ensures the APIs are properly exposed where relevant.

In a follow up PR, I am going to take the friendly wrappers which try to mirror the Clang C++ Type hierarchy (i.e. the files under [https://github.com/microsoft/ClangSharp/tree/master/sources/ClangSharp.PInvokeGenerator/Cursors](sources/ClangSharp.PInvokeGenerator/Cursors) and [https://github.com/microsoft/ClangSharp/tree/master/sources/ClangSharp.PInvokeGenerator/Types](sources/ClangSharp.PInvokeGenerator/Types)) and move them into ClangSharp proper.
-- This will involve making them public and ensuring the appropriate checks exist where needed.


tannergooding:
After the follow up PR I mentioned above, I think we should be good to publish an updated preview package for both `ClangSharp` and `ClangSharpPInvokeGenerator` (the latter of which I want to do as a global tool: https://github.com/microsoft/ClangSharp/issues/46).

This will basically just give you:
* If you import `ClangSharp`, you get the friendly/safe object oriented wrapper that should "just work"
* If you import `ClangSharp.Interop`, you have your choice between the raw unsafe API, and the simple abstraction layer that removes the "unsafe" bits.

tannergooding:
It's also worth noting that there are `5` APIs which aren't exposed in the "simple wrapper". This is because they are more "global" methods and don't operate on any wrapper type.

These are:
* `enableStackTraces`
* `executeOnThread`
* `getClangVersion`
* `toggleCrashRecovery`
* `getBuildSessionTimestamp`

PR #73, opened 2019-06-16
Expose the Cursor/Type wrappers as part of ClangSharp

tannergooding:
This is a follow up to https://github.com/microsoft/ClangSharp/pull/72 and moves the `Cursor`/`Type` wrappers to be part of `ClangSharp` proper.

It does some various cleanup to ensure that the appropriate validation is being done and to ensure that any cached values are cleaned up.


sources/ClangSharp/Cursors/Decls/AccessSpecDecl.cs
```diff
@@ -0,0 +1,11 @@
+using ClangSharp.Interop;
+
+namespace ClangSharp
+{
+    public sealed class AccessSpecDecl : Decl
+    {
+        internal AccessSpecDecl(CXCursor handle) : base(handle, CXCursorKind.CXCursor_CXXAccessSpecifier)
```
tannergooding: Most of the types were copy pasted and had the accessibility of the constructor fixed up.

The `Debug.Assert(handle.Kind == <expectedKind>)` calls were changed to pass the expected kind down to the base constructor.

sources/ClangSharp/Cursors/Decls/CXXRecordDecl.cs
```diff
@@ -0,0 +1,41 @@
...
+        private readonly Lazy<IReadOnlyList<CXXBaseSpecifier>> _bases;
+        private readonly Lazy<IReadOnlyList<CXXConstructorDecl>> _ctors;
+        private readonly Lazy<CXXDestructorDecl> _destructor;
+        private readonly Lazy<IReadOnlyList<FriendDecl>> _friends;
+        private readonly Lazy<IReadOnlyList<CXXMethodDecl>> _methods;
+        private readonly Lazy<IReadOnlyList<CXXBaseSpecifier>> _vbases;
+
+        internal CXXRecordDecl(CXCursor handle, CXCursorKind expectedKind) : base(handle, expectedKind)
+        {
+            _bases = new Lazy<IReadOnlyList<CXXBaseSpecifier>>(() => CursorChildren.Where((cursor) => cursor is CXXBaseSpecifier).Cast<CXXBaseSpecifier>().ToList());
```
tannergooding: Most of the traversal logic was changed to be lazy and to just pull any specific exposures out via caching the result of a LINQ operation.

sources/ClangSharp/Cursors/Decls/TagDecl.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using ClangSharp.Interop;
+
+namespace ClangSharp
+{
+    public unsafe class TagDecl : TypeDecl, IDeclContext, IRedeclarable<TagDecl>
```
tannergooding: There are interfaces matching the C++ type hierarchy for `DeclContext`, `IRedeclarable`, and `IMergeable`. The latter two are currently just marker interfaces (effectively), but I think I can expose the relevant semantics properly in a later PR.

PR #74, opened 2019-06-21
Minor improvements for type handling.

tannergooding:
This updates the PInvokeGenerator to:
* Handle attributed types, which frequently comes up for function pointers in some Windows headers
* Check for remapped names when processing typedefs. This allows users to handle things like `SIZE_T`, in the same manner as we were handling `size_t`.
* Adds a config switch for generating "unix-sized types". So, for example, `long` will emit `IntPtr` rather than `int`
* Adds "reverse" config switches so users can explicitly specify the behavior they want (last specified switch wins, for a given pair)
* Defaults to generating "unix-sized types" on Unix and "windows-sized types" on Windows

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
@@ -7,6 +7,7 @@ namespace ClangSharp
     {
         private const string DefaultMethodClassName = "Methods";
 
+        private readonly Dictionary<string, string> _remappedNames;
         private readonly PInvokeGeneratorConfigurationOptions _options;
 
         public PInvokeGeneratorConfiguration(string libraryPath, string namespaceName, string outputLocation, PInvokeGeneratorConfigurationOptions options = PInvokeGeneratorConfigurationOptions.None, string[] excludedNames = null, string methodClassName = null, string methodPrefixToStrip = null, IReadOnlyDictionary<string, string> remappedNames = null)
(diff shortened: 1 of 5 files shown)


tannergooding:
`-c unix-types` shows that `CXTUResourceUsageEntry` and `CXUnsavedFile` both have potential problems.

The former has `unsigned long amount` and the latter `unsigned long Length`, so you need to cross-compile to properly support both Unix and Windows.

tannergooding:
For LLVM, the same shows `off_t` resolves to be `long` on Windows which impacts `lto_module_create_from_fd_at_offset`

sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
```diff
@@ -54,13 +50,32 @@ public PInvokeGeneratorConfiguration(string libraryPath, string namespaceName, s
             MethodPrefixToStrip = methodPrefixToStrip;
             Namespace = namespaceName;
             OutputLocation = outputLocation;
-            RemappedNames = remappedNames;
+
+            _remappedNames = new Dictionary<string, string>()
```
mjsabby: can this be made config?

tannergooding: The current hardcoded list represents 4 types defined by the C/C++ standard to be platform sized, so this gives you the correct behavior by default for common scenarios.

You can also override these with new values if you wanted to customize them. -- The only thing you can't do is resolve them to the base typedef via "normal" logic, which is what a switch would allow.

I can certainly add a switch (maybe `no-default-remap` or something).

mjsabby:
LGTM

PR #75, opened 2019-06-23
Some more minor improvements

tannergooding:
* Adds a config switch to ignore any default remappings added.
* Resolves the full path of `-o` so that `-o file.cs` works (it was creating a directory with that name currently)
* Adds a `--traverse` option so that you can `-f header.cs` and then generate bindings for a nested include
  * This is useful when you need to generate bindings for a header file which isn't "standalone"
* Fixes `CXUnsavedFile` and `CXTUResourceUsageEntry` to use `uint` on Windows and `UIntPtr` on Unix for their `unsigned long` usages
  * Going to log an issue tracking a general purpose solution for this, so it isn't manual

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -175,9 +175,22 @@ namespace ClangSharp
 
             foreach (var decl in translationUnitDecl.Decls)
             {
-                if (!decl.IsFromMainFile)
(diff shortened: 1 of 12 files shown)


tannergooding:
Also pushed up a commit which simplifies how the opcode for a binary/unary operator is computed.

The logic in `BinaryOperator` specifically wasn't accounting for the fact that `LHS` could itself be a non-parenthesized binary expression itself.

Now, it just grabs the first opcode after the LHS expression, which should always be the opcode.

PR #80, opened 2019-08-08
Always write headers and check against empty paths

### LINKED ISSUE #78: Not specifying output header file (-h) casues ArgumentException
In PInvokeGeneratorConfiguration, headerFile seems to be null-checked and passed to `File.ReadAllText`:

https://github.com/microsoft/ClangSharp/blob/b15d4a79facfec91782abfec91236fe031c2c9df/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs#L54

However, the headerFile option defaults to `string.Empty`, meaning this check always succeeds:

https://github.com/microsoft/ClangSharp/blob/b15d4a79facfec91782abfec91236fe031c2c9df/sources/ClangSharpPInvokeGenerator/Program.cs#L329-L337

This causes the following exception if the parameter is not specified, when running the program:
```
System.ArgumentException: Empty path name is not legal.
Parameter name: path
   at System.IO.File.ReadAllText(String path)
   at ClangSharp.PInvokeGeneratorConfiguration..ctor(String libraryPath, String namespaceName, String outputLocation, PInvokeGeneratorConfigurationOptions options, String[] excludedNames, String headerFile, String methodClassName, String methodPrefixToStrip, IReadOnlyDictionary`2 remappedNames, String[] traversalNames) in /home/dev/Development/ClangSharp/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs:line 54
   at ClangSharp.Program.Run(InvocationContext context) in /home/dev/Development/ClangSharp/sources/ClangSharpPInvokeGenerator/Program.cs:line 173
```

### LINKED ISSUE #79: Output header file is not used for generated types in multi-file mode
When using `--config multi-file`, ClangSharpPInvokeGenerator does not use the header file (`-h`) when generating files for subtypes - only the main methods file gets the header file.

From cursory inspection, this is probably because the generated subtypes do not have any usings, and the header text is only added if there are any using directives:

https://github.com/microsoft/ClangSharp/blob/b15d4a79facfec91782abfec91236fe031c2c9df/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs#L271-L273

FiniteReality:
This fixes #78 and fixes #79. Header files are written whenever a valid header file is passed, and null/empty paths are now ignored.

Sorry about the whitespace changes, I didn't realise I had VS Code set to hide whitespace changes in git diffs.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -268,20 +268,25 @@ namespace ClangSharp
             using var sw = new StreamWriter(stream, defaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen);
             sw.NewLine = "\n";
 
-            if (_config.HeaderText != string.Empty && !isMethodClass)
-                sw.WriteLine(_config.HeaderText);
-
-            if (outputBuilder.UsingDirectives.Any() && _config.GenerateMultipleFiles)
+            if (_config.GenerateMultipleFiles)
             {
-                foreach (var usingDirective in outputBuilder.UsingDirectives)
+                if (_config.HeaderText != string.Empty)
                 {
-                    sw.Write("using");
-                    sw.Write(' ');
-                    sw.Write(usingDirective);
-                    sw.WriteLine(';');
+                    sw.WriteLine(_config.HeaderText);
                 }
 
-                sw.WriteLine();
+                if (outputBuilder.UsingDirectives.Any())
+                {
+                    foreach (var usingDirective in outputBuilder.UsingDirectives)
+                    {
+                        sw.Write("using");
+                        sw.Write(' ');
+                        sw.Write(usingDirective);
+                        sw.WriteLine(';');
+                    }
+
+                    sw.WriteLine();
+                }
             }
 
             var indentationString = outputBuilder.IndentationString;
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -268,7 +268,7 @@ namespace ClangSharp
             using var sw = new StreamWriter(stream, defaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen);
             sw.NewLine = "\n";
 
-            if (_config.HeaderText != string.Empty)
+            if (_config.HeaderText != string.Empty && !isMethodClass)
                 sw.WriteLine(_config.HeaderText);
 
             if (outputBuilder.UsingDirectives.Any() && _config.GenerateMultipleFiles)
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -268,10 +268,11 @@ namespace ClangSharp
             using var sw = new StreamWriter(stream, defaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen);
             sw.NewLine = "\n";
 
+            if (_config.HeaderText != string.Empty)
+                sw.WriteLine(_config.HeaderText);
+
             if (outputBuilder.UsingDirectives.Any() && _config.GenerateMultipleFiles)
             {
-                sw.Write(_config.HeaderText);
-
                 foreach (var usingDirective in outputBuilder.UsingDirectives)
                 {
                     sw.Write("using");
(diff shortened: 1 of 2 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -1399,12 +1400,12 @@ private void VisitParmVarDecl(ParmVarDecl parmVarDecl, TypedefDecl typedefDecl)
             var parameters = typedefDecl.CursorChildren.Where((cursor) => cursor is ParmVarDecl).Cast<ParmVarDecl>().ToList();
             var index = parameters.IndexOf(parmVarDecl);
             var lastIndex = parameters.Count - 1;
-            
+
             if (name.Equals("param"))
             {
                 _outputBuilder.Write(index);
             }
-
```
tannergooding: 👍, the formatter should have removed this and it shouldn't have been committed in the first place

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -268,10 +268,11 @@ private void CloseOutputBuilder(Stream stream, OutputBuilder outputBuilder, bool
             using var sw = new StreamWriter(stream, defaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen);
             sw.NewLine = "\n";
 
+            if (_config.HeaderText != string.Empty && !isMethodClass)
```
tannergooding: Where will it be handled for the `methodClass` if not here?

tannergooding: Also, won't https://github.com/microsoft/ClangSharp/issues/79 also apply to the single-file case if you have a header file without methods? So shouldn't https://github.com/microsoft/ClangSharp/blob/master/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs#L86 be fixed too?

Something like:
```cs
                using var sw = new StreamWriter(stream, defaultStreamWriterEncoding, DefaultStreamWriterBufferSize, leaveStreamOpen);
                {
                    sw.NewLine = "\n";
                    sw.Write(_config.HeaderText);

                    foreach (var usingDirective in usingDirectives)
                    {
                        sw.Write("using");
                        sw.Write(' ');
                        sw.Write(usingDirective);
                        sw.WriteLine(';');
                    }

                    sw.WriteLine();
                }
```

FiniteReality: From my tests this doesn't seem to happen, and the header always ends up getting written. I am not completely sure if I have tested every edge case though.

Maybe a better approach would be to change CloseOutputBuilder to accept an "additional usings" parameter and have it write everything, including the header?

tannergooding: Even in the case where there are no using statements for the single file?

tannergooding: Whats the output in that case?

Say if, for example, you just have a header file containing:
```cs
struct MyStruct
{
    int value;
}
```

tannergooding:
Test failures for windows look unrelated. Looks like something might've changed with how paths are processed....

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
tannergooding: Ok, looks like the exact case above is fine, but if you do:
```
#include <stdint.h>

struct MyStruct
{
    intptr_t value;
};
```

Then you get the license file printed twice.

tannergooding:
```cs
// Copyright
using System;

// Copyright

namespace LLVMSharp
{
    {
        [NativeTypeName("intptr_t")]
        public IntPtr value;
    }
}
```

tannergooding: This should probably be...
```cs
if (_config.GenerateMultipleFiles)
{
    if (_config.HeaderText != string.Empty)
    {
        sw.WriteLine(_config.HeaderText);
    }

    if (outputBuilder.UsingDirectives.Any()
    {
        ...
    }
}
```

tannergooding: (The `!isMethodClass` was also causing issues, as the license wasn't generating for that file)

FiniteReality: :+1:

tannergooding:
Thanks @FiniteReality

PR #101, opened 2019-11-20
Updating ClangSharpPInvokeGenerator to pack as a global tool

tannergooding:
This updates ClangSharpPInvokeGenerator to pack as a global tool. Noting that this requires `libClang` and `libClangSharp` to exist somewhere on the path already. This is a due to a limitation in dotnet global tools not easily allowing rid specific distributions and the `libClang` dependencies being quite large.


tannergooding:
I have a few more fixes to the generator I want to get up and then a new package can be published.

PR #123, opened 2020-04-15
Adds a 'preview-codegen' config option

tannergooding:
This adds a 'preview-codegen' config option that can be used to generate code compatible with the latest C# Preview Compiler. For example, it supports this PR allows it to use the `nint`/`nuint` types and `function pointers`.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -842,9 +842,35 @@ namespace ClangSharp
             {
                 name = GetTypeName(cursor, elaboratedType.NamedType, out var nativeNamedTypeName);
             }
-            else if (type is FunctionType)
+            else if (type is FunctionType functionType)
             {
-                name = "IntPtr";
+                if (_config.GeneratePreviewCode && (functionType is FunctionProtoType functionProtoType))
+                {
+                    var remappedName = GetRemappedName(name, cursor, tryRemapOperatorName: false);
+                    var callConv = GetCallingConventionName(cursor, functionType.CallConv, remappedName).ToLower();
+
+                    var nameBuilder = new StringBuilder();
+                    nameBuilder.Append("delegate");
+                    nameBuilder.Append('*');
+                    nameBuilder.Append(' ');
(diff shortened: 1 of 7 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -842,9 +842,35 @@ private string GetTypeName(Cursor cursor, Type type, out string nativeTypeName)
...
+                if (_config.GeneratePreviewCode && (functionType is FunctionProtoType functionProtoType))
+                {
+                    var remappedName = GetRemappedName(name, cursor, tryRemapOperatorName: false);
+                    var callConv = GetCallingConventionName(cursor, functionType.CallConv, remappedName).ToLower();
+
+                    var nameBuilder = new StringBuilder();
+                    nameBuilder.Append("delegate");
+                    nameBuilder.Append('*');
+                    nameBuilder.Append(' ');
+                    nameBuilder.Append((callConv != "winapi") ? callConv : "unmanaged");
```
333fred, line 856: Unmanaged is not currently supported, this will fail.

tannergooding: I might just keep this for now since its a preview feature and the code will get flagged in the compilation.
We have a few cases where the generated code isn't quite right and where devs may need to manually fixup or adjust the generated code.
They can workaround it by passing in the `--with-callconv RemappedName=CallConv` arg on the command line

333fred: Alright, but it's going to be weird when people get a CS8XXX for this.

tannergooding: I'll adjust if the language ends up not allowing the keyword in the final implementation. It also shouldn't be hit today unless they are targeting something using a non-standard calling convention.

This case will already print a `Unsupported calling convention` diagnostic since it isn't something we can support for `P/Invoke` calls or `Delegates w/ UnmanagedFunctionPointerAttribute`: https://github.com/microsoft/ClangSharp/blob/master/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs#L560

Issue #127, opened 2020-04-30
Investing providing the PInvoke Generator as a Source Generator


tannergooding:
As per https://devblogs.microsoft.com/dotnet/introducing-c-source-generators, C# now has support for Source Generators (in preview). Given that `Clang Sharp P/Invoke Binding Generator`'s whole job is to generate C# source code, we should likely investigate providing one as an alternative to the command line tool.

kkukshtel:
Bumping this to say it would be fantastic as something like the fin word on interop. If I could have .h files sitting in my main repo that get emitted to C# as part of a compile step, that would be incredible.

PR #133, opened 2020-05-05
Rewrite P/Invoke generator using Roslyn

FiniteReality:
**NOTE:** This is super *super* alpha; all I did was get a small sample application working before pushing these commits for now.

Rewriting the P/Invoke generator using Roslyn should allow us in the future to modify it to work with new features like source generators, which could make generating P/Invoke signatures *really* easy. (See #127)

However, no current public API has been properly designed or considered; what is here is currently the "bare minimum" in order to get something working. I do hope that this minimal example can solicit some dialogue, and hopefully eventually a final design.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -30,6 +30,8 @@
     <PackageReference Update="libClang" Version="10.0.0" />
     <PackageReference Update="libClangSharp" Version="10.0.0-beta1" />
     <PackageReference Update="Microsoft.Bcl.HashCode" Version="1.1.0" />
+    <PackageReference Update="Microsoft.CodeAnalysis" Version="3.5.0" />
+    <PackageReference Update="Microsoft.CodeAnalysis.CSharp" Version="3.5.0" />
     <PackageReference Update="Microsoft.Net.Compilers.Toolset" Version="3.5.0" />
     <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.5.0" />
     <PackageReference Update="System.CommandLine" Version="2.0.0-beta1.20158.1" />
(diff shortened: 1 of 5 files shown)


tannergooding:
Thanks @FiniteReality. I might take this and see if I can get a more complete sample working, is that alright with you?

FiniteReality:
Fine with me - the current "sample" I'm working with is some code I borrowed from the Doxygen documentation:
```
typedef unsigned long long some_type;
/*! \brief a normal member taking two arguments and returning an integer value.
 *  \param a an integer argument.
 *  \param s a constant character pointer.
 *  \sa Javadoc_Test()
 *  \sa ~Javadoc_Test()
 *  \sa testMeToo()
 *  \sa publicVar()
 *  \return The test results
 */
some_type testMe(some_type a,const char *s);
```

I'm then simply iterating through the function declarations and using PInvokeMethodBuilder on them.

PathogenDavid:
> Roslyn should allow us in the future to modify it to work with new features like source generators

While I don't doubt that this is a good idea in general, does this change actually accomplish this goal?

I am under the impression that source generators emit instances of `SourceText`, which are essentially unparsed strings rather than the syntax nodes emitted from `SyntaxFactory` as with this PR.

FiniteReality:
> While I don't doubt that this is a good idea in general, does this change actually accomplish this goal?

Not by itself, of course :stuck_out_tongue: 

> I am under the impression that source generators emit instances of SourceText, which are essentially unparsed strings rather than the syntax nodes emitted from SyntaxFactory as with this PR.

Yeah, but by returning a full syntax tree, we make it easier to integrate with extensions. For example, if you had an extension to ClangSharp which allowed it to properly read Qt signals and slots and translate those to events, exposing the full syntax tree makes it easier to integrate those with the P/Invoke generator, in my opinion.

PathogenDavid:
Roger that, thanks for the clarification!

PR #159, opened 2020-07-03
Extended the support for ClangSharp via libClangSharp

tannergooding:
The libClang bindings are very limited in what they expose and libClangSharp exists to help bridge that gap.

This expands libClangSharp to support quite a bit of additional functionality so the much of the guesswork no longer has to occur. Namely, the majority of the core properties on the C and C++ decls, statements, and expressions now have a way to directly query the relevant information. ObjC, OMP, and various Clang specific builtins do not currently have such support nor is absolutely everything exposed. However, this should make traversal of more complex ASTs easier and less error prone. It should also help with processing information out of the various template types.


sources/libClangSharp/ClangSharp.h
```diff
@@ -12,7 +14,7 @@
 #endif
 
 #ifdef _MSC_VER
-#ifdef CLANGSHARP_LINKAGE
+#ifdef clangsharp_LINKAGE
```
PathogenDavid, line 17: Shouldn't this be [`libClangSharp_EXPORTS`](https://cmake.org/cmake/help/v3.18/prop_tgt/DEFINE_SYMBOL.html)?

([`libClangSharp` because that's what it's called on Windows builds and this is for Windows.](https://github.com/microsoft/ClangSharp/blob/a42910cfcfd009880c2f12808b51f2dca25ba1c3/sources/libClangSharp/CMakeLists.txt#L41)

Better yet, can we just add our own preprocessor define instead of relying on the magic one added by CMake?

sources/libClangSharp/ClangSharp.cpp
```diff
@@ -54,80 +411,2151 @@ CX_CastKind clangsharp_Cursor_getCastKind(CXCursor C) {
...
+
+    return clang_getNullLocation();
+}
+
+CXCursor clangsharp_Cursor_getTemplateArgumentLocSourceDeclExpression(CXCursor C, unsigned i) {
+    if (clang_isDeclaration(C.kind)) {
+        const Decl* D = getCursorDecl(C);
+
+        if (const ClassScopeFunctionSpecializationDecl* CSFSD = dyn_cast<ClassScopeFunctionSpecializationDecl>(D)) {
+            const Expr* E = CSFSD->getTemplateArgsAsWritten()->arguments()[i].getSourceDeclExpression();
```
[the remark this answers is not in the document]
tannergooding: Most of libClang proper doesn't do bounds checking, so I didn't add it here either.

sources/libClangSharp/ClangSharp.h
tannergooding: no, `libclang` just uses `CLANG_LINKAGE`

PathogenDavid: I'm slightly confused now. Pretty sure `libclang` uses `CINDEX_LINKAGE` and that's not what this review comment is referring to.

I see you changed it back to `#ifdef CLANGSHARP_LINKAGE`, but I don't think that's right either.

```cpp
#ifdef CLANGSHARP_LINKAGE
#define CLANGSHARP_LINKAGE EXTERN_C __declspec(dllexport)
#else
#define CLANGSHARP_LINKAGE EXTERN_C __declspec(dllimport)
#endif
```

This can only ever do two things:

1) `CLANGSHARP_LINKAGE` will be defined twice, causing an C4005 warning.
2) `CLANGSHARP_LINKAGE` will defined as `EXTERN_C __declspec(dllimport)` (which isn't right when we're building a DLL.)

I feel like I'm missing something here since it shouldn't even build without `CLANGSHARP_LINKAGE` (and even with it defined there'll be a warning.) Maybe I'm blind, but it doesn't look like libClangSharp is even built by the build scripts at all. Are you just manually invoking cmake and creating the NuGet package for it manually?

sources/libClangSharp/ClangSharp.cpp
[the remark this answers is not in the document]
PathogenDavid: That seems reasonable then. It just stood out to me since the `i == 0` below is a bounds check of sorts (kinda but not really.)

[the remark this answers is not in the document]
tannergooding: Those ones are specifically because I didn't want to come up with a different name for the variants that only ever have a single value, the `i` in those cases is effectively ignored 😅

sources/libClangSharp/ClangSharp.h
tannergooding: It looks to be "working" because `CLANGSHARP_LINKAGE` isn't defined and therefore it is always exported. It ultimately doesn't matter because no one is going to use this from C code anyways, so most of it doesn't apply.
I will, however, fix it to just always export with a note that no one should be using this from another C library.

PathogenDavid: Sounds good

PR #172, opened 2020-09-15
Fix struct parsing in C mode

### LINKED ISSUE #171: Simple struct in C mode throws a nullref
The following piece of C code throws an exception somewhere in the pinvoke generation process.

```c
typedef struct
{
  int a;
} b;
```

mjsabby:
Fixes #171

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -1064,21 +1064,24 @@ namespace ClangSharp
                     }
                 }
 
-                foreach (var cxxConstructorDecl in cxxRecordDecl.Ctors)
+                if (cxxRecordDecl != null)
                 {
-                    Visit(cxxConstructorDecl);
-                    _outputBuilder.NeedsNewline = true;
-                }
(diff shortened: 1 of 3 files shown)


tannergooding:
👍, could we get a small regression test for this scenario?

You'll probably just need to tweak https://github.com/microsoft/ClangSharp/blob/master/tests/ClangSharp.PInvokeGenerator.UnitTests/PInvokeGeneratorTest.cs#L29 to take in the command line args, defaulting `null` to mean `DefaultClangCommandLineArgs`.

mjsabby:
@tannergooding oops, forgot to include it in the commit, it is as you recommended.

tannergooding (approved):

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -1096,7 +1099,7 @@ private void VisitRecordDecl(RecordDecl recordDecl)
                     VisitConstantArrayFieldDecl(recordDecl, constantArray);
                 }
 
-                if (hasVtbl)
+                if (cxxRecordDecl != null && hasVtbl)
```
tannergooding, line 1102: just a note, `hasVtbl` shouldn't ever be `true` if `cxxRecordDecl` is null; so this isn't "strictly" needed.

PR #174, opened 2020-09-16
Updating to use the new fnptr syntax and fixing a couple minor issues

tannergooding:
The alignment and callexpr issues were found when processing libX11 and libVulkan.


sotteson1:
I'm wondering about this change:

                _outputBuilder.WriteIndented("[UnmanagedFunctionPointer");

                if (callingConventionName != "Winapi")
                {
                    _outputBuilder.Write("(CallingConvention.");
                    _outputBuilder.Write(callingConventionName);
                    _outputBuilder.Write(')');
                }

                _outputBuilder.WriteLine(']');

That attribute seems to only have one constructor that takes the calling convention, but it's left out if the calling convention isn't Winapi. I'm sure I'm missing something, but I don't see how this should work if the calling convention is Winapi.

PR #176, opened 2020-09-16
Invalid enums should default to int

mjsabby:


tannergooding:
What was the code that was causing this and the error that it was raising?

PR #181, opened 2020-10-19
Regenerating bindings for Clang 11.0.0

tannergooding:
This updates ClangSharp to use libClang v11.0.0 and resolves https://github.com/microsoft/ClangSharp/issues/180


tannergooding:
This is pending building macOS binaries. This should happen around Wednesday.

PR #187, opened 2020-12-08
Adding new config options for ClangSharp

sotteson1:
Some changes required for a project where we use ClangSharp to against Win32 headers so that they can be projected into other languages.

- Adds 5 new config options:

1. ExcludeFunctionsWithBody: exclude functions that have a body (code). For projecting to other languages, we want to avoid any functions that include code. We only want DLL import functions.
2. ExcludeAnonymousFieldHelpers: exclude helper code for anonymous fields.
3. LogPotentialTypedefRemappings: adds diagnostics to suggest remappings (e.g. tagMyStruct to MyStruct).
4. GenerateCppAttributes: adds a CppAttributeList attribute for parameters that have C++ attributes on them. We are using this for SAL attributes which help us figure out, for example, if a pointer to a struct is just a single struct or an array.
5. GenerateNativeInheritanceAttribute: adds an attribute that shows the last base type the struct derived from. We use this to help figure out COM interface inheritance.

- Fixed a few issues we ran into:

1. An anonymous struct that wasn't a nested type was getting named "_Anonymous_e__Struct" but then was referenced in a function as the header/location name (e.g. __AnonymousRecord_myheader_L275_C9), so it wouldn't compile. Make sure to only rename to "_Anonymous_e__Struct" if it's a nested type.
2. This wasn't safe to cast in one case we saw: var functionType = (FunctionType)type; So instead use "is" to make sure it really is of type FunctionType.
3. When getting the type for an anonymous enum, get its integer type instead of its anonymous type name, since anonymous enum values are emitted as constants and not as constants in an enum type.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -345,7 +345,7 @@ namespace ClangSharp
 
             var type = fieldDecl.Type;
             var typeName = GetRemappedTypeName(fieldDecl, context: null, type, out var nativeTypeName);
-            
+
             if (fieldDecl.Parent.IsUnion)
             {
                 _outputBuilder.WriteIndentedLine("[FieldOffset(0)]");
@@ -1134,10 +1134,6 @@ namespace ClangSharp
                 {
                     OutputMethods(cxxRecordDecl, cxxRecordDecl);
                     excludedCursors = excludedCursors.Concat(cxxRecordDecl.Methods);
-                }
-
-                if (recordDecl.Spelling == "WhitePoint")
-                {
                 }
 
                 Visit(recordDecl.Decls, excludedCursors);
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -431,13 +431,11 @@ namespace ClangSharp
                 type = attributedType.ModifiedType;
                 callConv = attributedType.Handle.FunctionTypeCallingConv;
             }
+            var functionType = (FunctionType)type;
 
-            if (type is FunctionType functionType)
+            if (callConv == CXCallingConv.CXCallingConv_Invalid)
             {
-                if (callConv == CXCallingConv.CXCallingConv_Invalid)
-                {
-                    callConv = functionType.CallConv;
-                }
+                callConv = functionType.CallConv;
             }
 
             var cxxMethodDecl = functionDecl as CXXMethodDecl;
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -431,11 +431,13 @@ namespace ClangSharp
                 type = attributedType.ModifiedType;
                 callConv = attributedType.Handle.FunctionTypeCallingConv;
             }
-            var functionType = (FunctionType)type;
 
-            if (callConv == CXCallingConv.CXCallingConv_Invalid)
+            if (type is FunctionType functionType)
             {
-                callConv = functionType.CallConv;
+                if (callConv == CXCallingConv.CXCallingConv_Invalid)
+                {
+                    callConv = functionType.CallConv;
+                }
             }
 
             var cxxMethodDecl = functionDecl as CXXMethodDecl;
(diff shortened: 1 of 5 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -431,11 +431,13 @@ private void VisitFunctionDecl(FunctionDecl functionDecl)
                 type = attributedType.ModifiedType;
                 callConv = attributedType.Handle.FunctionTypeCallingConv;
             }
-            var functionType = (FunctionType)type;
 
-            if (callConv == CXCallingConv.CXCallingConv_Invalid)
+            if (type is FunctionType functionType)
             {
-                callConv = functionType.CallConv;
+                if (callConv == CXCallingConv.CXCallingConv_Invalid)
```
tannergooding, line 437: nit: Simplify the nesting level here
```suggestion
if ((type is FunctionType functionType) && (callConv == CXCallingConv.CXCallingConv_Invalid))
```

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -1590,6 +1595,11 @@ void VisitAnonymousRecordDecl(RecordDecl recordDecl, RecordDecl nestedRecordDecl
 
             void VisitAnonymousRecordDeclFields(RecordDecl rootRecordDecl, RecordDecl anonymousRecordDecl, string contextType, string contextName)
             {
+                if (_config.ExcludeAnonymousFieldHelpers)
```
tannergooding, line 1598: Do you think this should be a general `ExcludeHelpers` or that there should be an `Exclude*Helper` per "helper type" (just to cover any future helper additions)

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -2489,9 +2499,26 @@ void ForUnderlyingType(TypedefDecl typedefDecl, Type underlyingType)
                 {
                     ForPointeeType(typedefDecl, parentType: null, referenceType.PointeeType);
                 }
-                else if (underlyingType is TagType)
+                else if (underlyingType is TagType underlyingTagType)
                 {
-                    // Nothing to do for tag types
+                    // See if there's a potential typedef remapping we want to log
```
tannergooding, line 2504: nit: Just for consistency, I think this logic should be in a `ForTagType` local function. It can then early exit with `if (!_config.LogPotentialTypedefRemappings)`

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -2489,9 +2499,26 @@ void ForUnderlyingType(TypedefDecl typedefDecl, Type underlyingType)
...
+                    if (_config.LogPotentialTypedefRemappings)
+                    {
+                        var typedefName = typedefDecl.UnderlyingDecl.Name;
+                        var possibleNamesToRemap = new string[] { "_" + typedefName, "_tag" + typedefName, "tag" + typedefName };
```
tannergooding, line 2508: Just looking at the Windows bindings I've done for my own project, what about `$"{typedefName}_tag"`, `$"{typedefName}_"`, and `$"{typedefName}__"`?

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -295,6 +295,94 @@ private void AddDiagnostic(DiagnosticLevel level, string message, Cursor cursor)
             _diagnostics.Add(diagnostic);
         }
 
+        private void AddCppAttributes(ParmVarDecl parmVarDecl, string prefix = null, string postfix = null)
```
tannergooding, line 298: Just want to confirm. This basically saves all the C++ attributes separated by `^` in a C# style attribute?

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -1906,6 +1999,14 @@ private bool IsExcluded(Cursor cursor, out bool isExcludedByConflictingDefinitio
                 return false;
             }
 
+            if (_config.ExcludeFunctionsWithBody &&
```
tannergooding, line 2002: `IsExcluded` is actually one of the "hottest" functions in ClangSharp. It would probably be good to keep this after the `IsExcludedByFile` check which catches most of the exclusions.

tannergooding:
Changes LGTM overall. Just a couple questions/suggestions and I think this can be merged.

tannergooding (approved):

tannergooding:
Going to take this and will take care of the few nits in a follow up PR.

PR #188, opened 2021-01-05
Fix translation for implicit move constructor invocation

reflectronic:
Move constructors need the same special handling as copy constructors do. 

One scenario that this fixes is functions which return structs. Previously, for C++ code like this: 
```cpp
struct MyStruct
{
    double r;
    double g;
    double b;
};

MyStruct MyFunction()
{
    MyStruct myStruct;
    return myStruct;
}
```
the `MyFunction` function would be translated as such: 
```cs
public static MyStruct MyFunction()
{
    MyStruct myStruct = new MyStruct();
    return new MyStruct(myStruct);
}
```
This patch removes the explicit constructor invocation.


tannergooding (approved):

tannergooding:
Talked with @reflectronic offline and while ideally this would check if the constructors are implicit, they can't due to lack of expose surface area today.
I'll log an issue to get that exposed and to update this logic.

PR #192, opened 2021-01-10
Add support for templates in type signatures

reflectronic:
Very rudimentary template support is present. This doesn't add support for template type declarations, only the ability to translate elements like `MyStruct<int>` in type signatures. Unfortunately, since pointers in generics aren't yet allowed in .NET, any pointer types in template arguments get erased uniformly as `UIntPtr` when translating to C#.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -1027,6 +1027,10 @@ namespace ClangSharp
                 {
                     baseType = elaboratedType.CanonicalType;
                 }
+                else if (baseType is TemplateSpecializationType templateSpecializationType)
+                {
+                    baseType = templateSpecializationType.CanonicalType;
+                }
                 else if (baseType is TypedefType typedefType)
                 {
                     baseType = typedefType.Decl.UnderlyingType;
@@ -1139,12 +1143,6 @@ namespace ClangSharp
         {
             var name = GetTypeName(cursor, context, type, out nativeTypeName);
             name = GetRemappedName(name, cursor, tryRemapOperatorName: false);
-
-            if (name.Contains("::"))
-            {
-                name = name.Split(new string[] { "::" }, StringSplitOptions.RemoveEmptyEntries).Last();
-                name = GetRemappedName(name, cursor, tryRemapOperatorName: false);
-            }
 
             var canonicalType = type.CanonicalType;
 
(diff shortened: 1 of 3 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -1389,6 +1386,50 @@ private string GetTypeName(Cursor cursor, Cursor context, Type type, out string
...
+
+                bool shouldWritePrecedingComma = false;
+
+                foreach (var arg in templateSpecializationType.Args)
+                {
+                    if (shouldWritePrecedingComma)
+                    {
+                        nameBuilder.Append(',');
+                        nameBuilder.Append(' ');
+
```
tannergooding, line 1406:
```suggestion
```

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -1389,6 +1386,50 @@ private string GetTypeName(Cursor cursor, Cursor context, Type type, out string
...
+                    if (typeName == "bool")
+                    {
+                        // bool is not blittable, so we shouldn't use it for P/Invoke signatures
+                        typeName = "byte";
+                    }
+
+                    if (typeName.EndsWith("*"))
+                    {
+                        // Pointers are not yet supported as generic arguments; remap to UIntPtr
+                        typeName = "UIntPtr";
```
tannergooding, line 1420: Should this account for "latest" vs "compatible" codegen and use `nuint`?

tests/ClangSharp.PInvokeGenerator.UnitTests/FunctionDeclarationDllImportTest.cs
```diff
@@ -69,6 +69,32 @@ public static partial class Methods
             await ValidateGeneratedBindingsAsync(inputContents, expectedOutputContents);
         }
 
+        [Theory]
+        [InlineData("MyStruct<int>", @"MyStruct<int>", "")]
+        [InlineData("MyStruct<bool>", @"[NativeTypeName(""MyStruct<bool>"")] MyStruct<byte>", "")]
```
tannergooding, line 74: What if the type appears in a non P/Invoke signature?

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
reflectronic: This `UIntPtr` is semantically equal to a `void*`, which doesn't provide any operations (like addition or subtraction) unless you cast to a different type of pointer. I don't think `nuint` would be an appropriate analog for that.

tannergooding: In that case, most .NET handles are exposed as `IntPtr` and that is what most of the bindings in libraries like TerraFX or WinForms use.

tannergooding: There also isn't a lot of difference, realistically speaking, between `UIntPtr` and `nuint`. Basically the only difference is that `nuint` supports the full set of operations and not just addition/subtraction with an `int`.

reflectronic: I think `nuint` just makes less sense. It implies you are supposed to perform operations on the raw integer value (when you really shouldn't), and hides the idea the value is actually just a (type erased) reference to a memory location. If think this should be `IntPtr` instead, that is okay; however I figured that, since these are actual memory addresses, and not simply opaque handles, that the explicit unsigned makes more sense.

tests/ClangSharp.PInvokeGenerator.UnitTests/FunctionDeclarationDllImportTest.cs
reflectronic: Added a test for this.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
tannergooding: Could you get an estimate of how much API surface would change in ClangSharp itself as well as something like TerraFX with the change to use `UIntPtr` as the default fallback type for pointers?

I'm namely concerned about this change: https://github.com/microsoft/ClangSharp/pull/192/files#diff-6a1f236ee26a1c65a174696f8f83527ae8fa16e6570aeedec19daec3ca6d70b4R1590 having widespread downstream impact and possibly usability concerns with .NET in general.

There is somewhat of a similar concern with representing `char` as `sbyte` by default considering that most of the UTF8 related APIs in .NET operate on `byte` instead, even though `sbyte` is more accurate to the MSVC/Clang defaults.

reflectronic: Okay, I reverted the aforementioned change and this remaps to `IntPtr` now.

tannergooding (approved):

PR #194, opened 2021-01-11
Add tests for anonymous enums

reflectronic:
Do not use the mangled anonymous name in `using static` directives. Additionally, a `using static` directive should not be added when members of the enum are already available in context (for example, when inside the method class.)

### DIFF
--- a/tests/ClangSharp.PInvokeGenerator.UnitTests/Base/EnumDeclarationTest.cs
+++ b/tests/ClangSharp.PInvokeGenerator.UnitTests/Base/EnumDeclarationTest.cs
@@ -63,6 +63,12 @@ public abstract class EnumDeclarationTest : PInvokeGeneratorTest
     [Test]
     public Task WithTypeStarOverrideTest() => WithTypeStarOverrideTestImpl();
 
+    [Test]
+    public Task WithAnonymousEnumTest() => WithAnonymousEnumTestImpl();
+
+    [Test]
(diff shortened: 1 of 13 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -391,9 +391,10 @@ private void VisitCXXUuidofExpr(CXXUuidofExpr cxxUuidofExpr)
 
         private void VisitDeclRefExpr(DeclRefExpr declRefExpr)
         {
-            if ((declRefExpr.Decl is EnumConstantDecl enumConstantDecl) && (declRefExpr.DeclContext != enumConstantDecl.DeclContext) && (enumConstantDecl.DeclContext is NamedDecl namedDecl))
+            if ((declRefExpr.Decl is EnumConstantDecl { DeclContext: NamedDecl namedDecl } enumConstantDecl) &&
```
tannergooding, line 394: nit: I really don't think this pattern is more readable than the `&&` expression.
It's likely a point in time thing due to patterns being less widespread, but at least right now, this is much harder for me to parse and understand what is being checked.

tannergooding:
@reflectronic are you interested in taking this forward still?

reflectronic:
Looks like the tests I added are passing in main as-is, so I'll just update this PR to add these tests (anonymous enums seem to be untested currently).

PR #197, opened 2021-01-30
Insert cast to enum underlying type when casting to enum type in enum initializer

reflectronic:
In C#, generally, casts to enum types inside of enum initializers cause things to stop working. For example:
```cs
public enum E1
{
    F = 3
}

public enum E2
{
    M = E1.F + E1.F, // Compiles fine
    E = (E1) E1.F + E1.F // Does not work!
}
```

Removing all casts to enum types when inside an enum initializer avoids these issues.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
@@ -454,6 +454,15 @@ namespace ClangSharp
 
         private void VisitExplicitCastExpr(ExplicitCastExpr explicitCastExpr)
         {
+            if (IsPrevContextDecl<EnumConstantDecl>(out var _) && explicitCastExpr.Type is EnumType)
+            {
+                // In C#, explicitly casting to an enum type in an enum initializer usually results in compile errors.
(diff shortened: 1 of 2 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -454,6 +454,15 @@ private void VisitDoStmt(DoStmt doStmt)
 
         private void VisitExplicitCastExpr(ExplicitCastExpr explicitCastExpr)
         {
+            if (IsPrevContextDecl<EnumConstantDecl>(out var _) && explicitCastExpr.Type is EnumType)
```
tannergooding, line 457: I don't think removing the cast is going to be correct in all cases
imagine a scenario where the value is `-1` and the target enum base type is `long` or `uint`

It's likely "safer" to insert an additional cast to the base type instead or possibly to replace it with a cast to the target enum base type

reflectronic: Okay, it will insert a cast to the underlying type instead now.

tannergooding (approved):

PR #198, opened 2021-01-30
Add a CI job for validating microsoft/win32metadata hasn't regressed

### LINKED ISSUE #195: Add a CI test to validate win32metadata is not broken
Now that microsoft/win32metadata is public, it would be good to add a CI leg to ClangSharp to ensure any new PRs don't regress or break win32metadata by default.

tannergooding:
This resolves #195 and depends on https://github.com/microsoft/win32metadata/pull/200

### DIFF
--- a/scripts/azure-pipelines.yml
+++ b/scripts/azure-pipelines.yml
@@ -11,6 +11,7 @@ jobs:
     pool: windows-latest
     configuration: Debug
     architecture: x86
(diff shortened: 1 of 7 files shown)


tannergooding:
CC. @sotteson1, it looks like the only flag that was missing is `dont-use-using-statics-for-enums`

PR #199, opened 2021-01-30
Translate temporary object expressions as constructor invocations

reflectronic:
Translate 
```cpp
X create_X() 
{
    return X(1, 3.14f);
}
```
 as 
```cs
X create_X() 
{
    return new X(1, 3.14f);
}


tannergooding:

PR #205, opened 2021-02-14
Support var decls in namespaces

reflectronic:


hermann-noll:
What would be the status of this PR?
The test needs a rebase but the CI error looks more like an internal CI error so is maybe already fixed?

tannergooding:

PR #206, opened 2021-02-28
Abstract away output mechanisms & add support for XML dumps

Perksey:


sources/ClangSharp.PInvokeGenerator/Abstractions/FunctionOrDelegateDesc.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+
+namespace ClangSharp.Abstractions
+{
+    public struct FunctionOrDelegateDesc<TCustomAttrGeneratorData>
+    {
+        public string AccessSpecifier;
```
tannergooding, line 7: I wonder if we could track some of these via enums, both to compress space (the swath of `bool` below) and to make things like `AccessSpecifier` and `CallingConvention` less tied to what C# expects the string to be and instead something more generic that can be reused across languages.

tannergooding: Would also be good to follow "best practices" for these and make the fields private and the exposed members properties with get/set (most can be auto, just may not the `bool` flags since those should be backed by a single `int32` or `int64`)

sources/ClangSharp.PInvokeGenerator/ClangSharp.PInvokeGenerator.csproj
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <RootNamespace>ClangSharp</RootNamespace>
     <TargetFrameworks>net5.0;netcoreapp3.1;netstandard2.0</TargetFrameworks>
+    <LangVersion>9</LangVersion>
```
tannergooding, line 7: nit: We can set this in the `LangVersion` in the root `Directory.Build.props`

sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorOutputMode.cs
```diff
@@ -0,0 +1,8 @@
+namespace ClangSharp
+{
+    public enum PInvokeGeneratorOutputMode
+    {
+        Csharp,
```
tannergooding, line 5: nit: `CSharp`

Perksey: Fully agree, just wasn't sure whether it had any implications from the System.CommandLine perspective

sources/ClangSharp.PInvokeGenerator/Abstractions/FunctionOrDelegateDesc.cs
```diff
@@ -16,83 +16,83 @@ public bool IsVirtual
         {
             get => (Flags & FunctionOrDelegateFlags.IsVirtual) != 0;
             set => Flags =
-                value ? Flags | FunctionOrDelegateFlags.IsVirtual : Flags ^ FunctionOrDelegateFlags.IsVirtual;
+                value ? Flags | FunctionOrDelegateFlags.IsVirtual : Flags & FunctionOrDelegateFlags.IsVirtual;
```
tannergooding: The false case needs to be `Flags & ~FunctionOrDelegateFlags.IsVirtual` so that it preserves all bits except for the one being cleared.

Perksey: Yeah I realised that, but left it on the college PC and forgot to push! Will fix tomorrow.

PR #212, opened 2021-03-21
Adding test coverage for the XML generator

tannergooding:


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -472,7 +472,7 @@ private void VisitFunctionDecl(FunctionDecl functionDecl)
 
             var needsReturnFixup = isVirtual && NeedsReturnFixup(cxxMethodDecl);
 
-            if (!(functionDecl is CXXConstructorDecl))
+            if ((functionDecl is not CXXConstructorDecl) || (_config.OutputMode == PInvokeGeneratorOutputMode.Xml))
```
Perksey, line 475: I guess we should add a boolean to WriteReturnType indicating whether it's for a ctor or not to avoid this being baked into VisitDecl, that does nothing if that boolean is true on the C# backend.

tests/ClangSharp.PInvokeGenerator.UnitTests/XmlCompatibleUnix/CXXMethodDeclarationTest.cs
```diff
@@ -0,0 +1,933 @@
...
+      </function>
+    </struct>
+  </namespace>
+</bindings>
+";
+
+            return ValidateGeneratedXmlCompatibleUnixBindingsAsync(inputContents, expectedOutputContents);
+        }
+
+        public override Task NewKeywordTest()
```
Perksey, line 356: I suppose this test is a bit redundant given we expect the downstream consumer to determine whether a new keyword is needed.

Perksey (commented):
Seems fine

PR #218, opened 2021-03-23
A couple fixes for the win32metadata project

sotteson1:
* If the config option is set not to output anonymous helper functions, leave them out.
* If we didn't find an expression in its parents args, look through the parent's args to see if any of them are an implicit cast that then contains the expression we didn't find. Without this, I get a crash because the index is -1.


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -1624,6 +1624,21 @@ private void VisitUnaryExprOrTypeTraitExpr(UnaryExprOrTypeTraitExpr unaryExprOrT
                         if (IsPrevContextStmt<CallExpr>(out var callExpr))
                         {
                             var index = callExpr.Args.IndexOf(unaryExprOrTypeTraitExpr);
+
+                            // If we didn't find the expression, try and find it under an implicit cast
+                            if (index == -1)
+                            {
+                                for (int i = 0; i < callExpr.Args.Count; i++)
```
tannergooding, line 1631: I wonder if it would be better to write this as a single loop:
```csharp
var args = callExpr.Args;
var index = -1;

for (int i = 0; i < args.Count; i++)
{
    var arg = args[i];

    if (IsStmtAsWritten(arg, unaryExprOrTypeTraitExpr))
    {
        index = i;
        break;
    }

    if (index == -1)
    {
        AddDiagnostic(DiagnosticLevel.Error, $"Failed to locate index of '{unaryExprOrTypeTraitExpr}' in callee declaration. Generated bindings may be incomplete.", calleeDecl);
    }
}
```

We can then have:
```csharp
private bool IsStmtAsWritten(Stmt stmt, Stmt expectedStmt, bool removeParens = false)
{
    if (stmt == expectedStmt)
    {
        return true;
    }

    if (stmt is not Expr expr)
    {
        return false;
    }

    expr = GetExprAsWritten(expr, removeParens);
    return expr == expectedStmt;
}
```

This will make it easier to cover more cases in the future and to consistently look into an expr to remove implicit casts or other bits as necessary.

tannergooding: We have an existing `IsStmtAsWritten` that does a type check, rather than a value check, and it follows roughly this pattern.

tannergooding:
> If we didn't find an expression in its parents args, look through the parent's args to see if any of them are an implicit cast that then contains the expression we didn't find. Without this, I get a crash because the index is -1.

It would be nice to have a small repro for this case. I'd like to add it and a test for `ExcludeAnonymousFieldHelpers` to ensure they don't break in the future. We have roughly 4 core modes we are currently testing:
* Windows vs Unix
* Compat vs Latest

tannergooding:
Going to merge this for now as it is correct as is and the tests/possible improvement to finding the `CallExpr` index can be done in a follow up PR.

sotteson1:
Thanks for the quick response @tannergooding!

tannergooding:
> Thanks for the quick response @tannergooding!

Sure thing @sotteson1! If you could confirm that things are working as expected for win32metdata now, then I can publish a new beta package to NuGet.

sotteson1:
> Sure thing @sotteson1! If you could confirm that things are working as expected for win32metdata now, then I can publish a new beta package to NuGet.

I'm using binaries built with the latest changes and things look great, so I'm ready try out a new package.

tannergooding:
An 11.0.0-beta3 package should now be available: https://www.nuget.org/packages/ClangSharpPInvokeGenerator/11.0.0-beta3

PR #221, opened 2021-03-28
Mark ternary macros that can be evaluated as constants as const

ThadHouse:
Previously, all conditionals were marked as not constant. However, if the Cond, LHS and RHS are all constant, the ternary can be marked constant as well

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -2861,7 +2861,8 @@ namespace ClangSharp
 
                 case CX_StmtClass.CX_StmtClass_ConditionalOperator:
                 {
-                    return false;
+                    var conditionalOperator = (ConditionalOperator)initExpr;
+                    return IsConstant(conditionalOperator.Cond) && IsConstant(conditionalOperator.LHS) && IsConstant(conditionalOperator.RHS);
                 }
 
                 // case CX_StmtClass.CX_StmtClass_AddrLabelExpr:
--- a/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/VarDeclarationTest.cs
+++ b/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/VarDeclarationTest.cs
@@ -300,5 +300,25 @@ namespace ClangSharp.Test
 
             return ValidateGeneratedCSharpCompatibleUnixBindingsAsync(inputContents, expectedOutputContents);
         }
+
+        public override Task ConditionalDefineConstTest()
+        {
+            var inputContents = @"typedef int TESTRESULT;
+#define TESTRESULT_FROM_WIN32(x) ((TESTRESULT)(x) <= 0 ? ((TESTRESULT)(x)) : ((TESTRESULT) (((x) & 0x0000FFFF) | (7 << 16) | 0x80000000)))
+#define ADDRESS_IN_USE TESTRESULT_FROM_WIN32(10048)";
+
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public static partial class Methods
+    {{
+        [NativeTypeName(""#define ADDRESS_IN_USE TESTRESULT_FROM_WIN32(10048)"")]
+        public const int ADDRESS_IN_USE = unchecked((int)(10048) <= 0 ? ((int)(10048)) : ((int)(((10048) & 0x0000FFFF) | (7 << 16) | 0x80000000)));
+    }}
+}}
+";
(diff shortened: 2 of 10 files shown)


tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/VarDeclarationTest.cs
```diff
@@ -300,5 +300,25 @@ public static partial class Methods
...
+            var expectedOutputContents = $@"namespace ClangSharp.Test
+{{
+    public static partial class Methods
+    {{
+        [NativeTypeName(""#define ADDRESS_IN_USE TESTRESULT_FROM_WIN32(10048)"")]
+        public const int ADDRESS_IN_USE = unchecked((int)(10048) <= 0 ? ((int)(10048)) : ((int)(((10048) & 0x0000FFFF) | (7 << 16) | 0x80000000)));
+    }}
+}}
+";
+            var diagnostics = new Diagnostic[] { new Diagnostic(DiagnosticLevel.Warning, "Function like macro definition records are not supported: 'TESTRESULT_FROM_WIN32'. Generated bindings may be incomplete.", "Line 2, Column 9 in ClangUnsavedFile.h") };
```
ThadHouse, line 319: Is there a better way to do this?

tannergooding: Not really, this is about how everything works for validating diagnostics.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -2861,7 +2861,8 @@ bool IsConstant(Expr initExpr)
 
                 case CX_StmtClass.CX_StmtClass_ConditionalOperator:
                 {
-                    return false;
+                    var conditionalOperator = (ConditionalOperator)initExpr;
+                    return IsConstant(conditionalOperator.Cond) && IsConstant(conditionalOperator.LHS) && IsConstant(conditionalOperator.RHS);
```
tannergooding, line 2865: nit: This isn't quite right. If `IsConstant(conditionalOperator.Cond)` is `true`, then only `IsConstant(conditionalOperator.LHS)` needs to be `true`.

ThadHouse: `a ? b :c` does that mean `Cond` is `a` and `LHS` is `b : c`? Otherwise confused by why only 2 of the 3 would need to be true.

ThadHouse: This code doesn't check that `Cond` evaluates to true. It just checks that it evaluates to a constant.

tannergooding: `cond ? LHS : RHS`

`x = cond ? LHS : RHS` is basically shorthand for:
```
if (cond)
{
    x = LHS;
}
else
{
    x = RHS;
}
```

So if you have `x = trueCond ? 5 : sqrt(globalValue)`, then you want it to still evaluate to be "constant". As it is now, it won't since `IsConstant(conditionalOperator.RHS)` will return `false`

tannergooding: Likewise if you had `x = falseCond ? sqrt(globalValue) : 5`, you want it to evaluate as constant while `IsConstant(conditionalOperator.LHS)` will be the only thing to return `false`

ThadHouse: Since the generator actually generates the ternary even if the `Cond` is constant, wouldn't the current impl need both LHS and RHS to be true? I agree if the generator actually lowered the ternary then only the evaluated condition would need to be checked, but thats not what gets generated currently.

tannergooding: Looks like yes actually. I thought C# would drop the dead code path here, but looks like its not.

ThadHouse: Yeah. I agree if the generator is changed to lower everything, then this evaluation could be more fine grained. But for now, it will generate the whole thing, so all 3 need to be constant evaluated in order to generate const.

tannergooding: Thanks, this looks good and I'll merge once CI finishes

ThadHouse:
Anything we can do to poke the windows build thats been stuck for a few weeks?

tannergooding:
It actually has passed and the status never got updated on the GitHub side. I'll force merge.

tannergooding (approved):

PR #227, opened 2021-04-06
Introduce EnumDesc, add ability to output source location attribute

andrew-boyarshin:
Not currently exposed, but can easily be plumbed as opt-in command line option. Note, that not all Desc items have Location attribute set, since for some it doesn't make any sense.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -300,6 +300,8 @@ namespace ClangSharp.CSharp
 
             if (!inline)
                 WriteNewline();
+            else
+                Write(' ');
         }
 
         public void WriteReturnType(string typeString)
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
 using ClangSharp.Abstractions;
+using ClangSharp.Interop;
 
 namespace ClangSharp.CSharp
 {
@@ -28,6 +29,9 @@ namespace ClangSharp.CSharp
             {
                 AddNativeTypeNameAttribute(desc.NativeTypeName);
             }
+
+            if (desc.Location is {} location)
+                WriteSourceLocation(location, false);
 
             WriteIndentation();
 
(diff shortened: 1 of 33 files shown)


sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
```diff
@@ -268,6 +281,27 @@ public void BeginFunctionOrDelegate<TCustomAttrGeneratorData>(in FunctionOrDeleg
             }
         }
 
+        private void WriteSourceLocation(CXSourceLocation location, bool inline)
```
tannergooding, line 284: I'm concerned this is going to end up very noisy across regenerations. The anonymous struct names/remappings is already a pain point for some users

tannergooding:
The rest of the changes look reasonable to me.

sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
andrew-boyarshin: It's opt-in. It is useful for diagnostics (aka "where did this come from?"), but not so much for end-user of the bindings. I don't expect SharpGen-generated bindings or e.g. TerraFX.Interop to contain this info. But e.g. win32metadata and SharpGen might use this information to enrich logging or other forms of diagnostics.

tannergooding: Right, I understand the benefit.

I was just iterating the concern around noise in practice. Many diffs would become nearly unreadable due to minor changes in information such as whitespace addition.

If it's being added here, a corresponding switch and basic test validating it works and doesn't regress should also be added.

andrew-boyarshin: Done. I've added 2 tests that cover all important cases.

sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfigurationOptions.cs
```diff
@@ -58,5 +58,7 @@ public enum PInvokeGeneratorConfigurationOptions
         GeneratePreviewCode = 0x00800000,
 
         GenerateTemplateBindings = 0x01000000,
+
+        GenerateSourceLocationAttribute = 1 << 25,
```
tannergooding, line 62: nit: This should be either `0x02000000` or all entries should be updated to use the `1 << n` syntax.

I don't have a particular preference for which is done, I'd just like for it to be consistent

tannergooding:
Changes LGTM, minus the differing flag value

sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfigurationOptions.cs
andrew-boyarshin: I'll update the rest to use bit shifting to be explicit and simplify visual validation of flag values.

andrew-boyarshin: Done.

tannergooding (approved):

PR #228, opened 2021-04-11
Provide more information in Desc structures

andrew-boyarshin:
Some changes to provide more info to the `IOutputBuilder` and enable generating e.g. additional attributes to be parsed out later.

Drop the only use of OutputMode in PInvokeGenerator binding generation code. This is no place for hardcoding `IOutputBuilder`-specific logic.


Perksey (approved):
Tysm for removing the hardcoded OutputMode! This was something that was added post-me that irked me a bit but apparently not enough to do anything about it...

tannergooding:
Sorry for the delay in getting this reviewed/merged.

Could you rebase or merge this with the latest main so we can get updated CI results?

PR #243, opened 2021-06-25
Handle virtual destructors

### LINKED ISSUE #240: Assertion fails at virtual destructor declaration
The tool crashes on failed assertion when there is a virtual destructor in a C++ class declaration:

ClangSharpPInvokeGenerator: /usr/lib/llvm/12/include/clang/AST/GlobalDecl.h:61: void clang::GlobalDecl::Init(const clang::Decl*): Assertion `!isa<CXXDestructorDecl>(D) && "Use other ctor with dtor decls!"' failed.

clang/llvm version 12.0.0, .NET Core 5.0

Example C++ file:
```c++
#ifndef INC_GUARD
#define INC_GUARD

struct IFace {
    virtual ~IFace() = 0;
}; 

#endif
```

zeule:
This resolves crashing on virtual destructors if Clang is build with
assertions and corrects vtbl indices for tables with destructors.

Fixes #240.

### DIFF
--- a/sources/libClangSharp/ClangSharp.cpp
+++ b/sources/libClangSharp/ClangSharp.cpp
@@ -8,12 +8,16 @@
 #include "CXTranslationUnit.h"
 #include "CXType.h"
 
+#ifdef _MSC_VER
 #pragma warning(push)
 #pragma warning(disable : 4146 4244 4267 4291 4624 4996)
+#endif
 
 #include <clang/Basic/SourceManager.h>
 
+#ifdef _MSC_VER
 #pragma warning(pop)
+#endif
 
 using namespace clang;
 using namespace clang::cxcursor;
@@ -59,6 +63,25 @@ bool isDeclOrTU(CXCursorKind kind) {
 
 bool isStmtOrExpr(CXCursorKind kind) {
     return clang_isStatement(kind) || clang_isExpression(kind);
+}
+
+int64_t getVtblIdx(const GlobalDecl& d)
+{
+    const CXXMethodDecl* CMD = static_cast<const CXXMethodDecl*>(d.getDecl());
+    if (VTableContextBase::hasVtableSlot(CMD)) {
+        VTableContextBase* VTC = CMD->getASTContext().getVTableContext();
+
+        if (MicrosoftVTableContext* MSVTC = dyn_cast<MicrosoftVTableContext>(VTC)) {
+            MethodVFTableLocation ML = MSVTC->getMethodVFTableLocation(d);
+            return ML.Index;
+        }
+
+        if (ItaniumVTableContext* IVTC = dyn_cast<ItaniumVTableContext>(VTC)) {
+            return IVTC->getMethodVTableIndex(d);
+        }
+    }
+
+    return -1;
 }
 
 CXCursor clangsharp_Cursor_getArgument(CXCursor C, unsigned i) {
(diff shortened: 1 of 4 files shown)


sources/libClangSharp/ClangSharp.cpp
```diff
@@ -4701,13 +4701,16 @@ int64_t clangsharp_Cursor_getVtblIdx(CXCursor C) {
             if (VTableContextBase::hasVtableSlot(CMD)) {
                 VTableContextBase* VTC = getASTUnit(getCursorTU(C))->getASTContext().getVTableContext();
 
+                const CXXDestructorDecl* dCMD = dyn_cast<CXXDestructorDecl>(CMD);
+                const GlobalDecl globalDecl = dCMD ? GlobalDecl(dCMD, Dtor_Deleting) : GlobalDecl(CMD);
```
tannergooding, line 4705: Looks like this also needs handling for constructors. `CUDAGlobalAttr` looks to be covered implicitly.

I also don't think this is quite the right fix. I think it would be better for this to be extended to also accept `GlobalDecl` directly and to explicitly have `getDtorVtblIdx` and `getCtorVtblIdx` that users should call so they can override which `DtorType`/`CtorType` they attempt to get the VTBL index for, given they may not always want the deleting dtor.

This makes it more extensible for the future while also allowing something that "works by default".

zeule: > Looks like this also needs handling for constructors…

Pardon, can't understand that. This block handles virtual functions, right? There can't be constructors here.

tannergooding: Ah right, the `hasVtableSlot` check will prevent this from being called for constructors. Nevermind on that part.

zeule: OK, thanks! Now I'm puzzled with the next one: virtual destructor declaration always creates the deleting destructor. What destructor type user can possibly select here?

tannergooding: Its been a while since I've had to read the rules here, but as far as I remember:

Declaring a virtual destructor will cause a pair of entries to be created, a complete and a deleting version. The complete version is simply the destruction logic (that is `~MyClass()`) while the deleting version is a specially generated version that also calls the relevant `delete` operator after invoking the destruction logic. This means that there are actually 2 virtual entries that may exist and which is called can be dependent on the scenario and whether calling `delete` is correct or not.

Its entirely possible I'm misremembering and conflating the concepts with how the VTBL is actually laid out however.

tannergooding: From https://itanium-cxx-abi.github.io/cxx-abi/abi.html#vtable
> The entries for virtual destructors are actually pairs of entries. The first destructor, called the complete object destructor, performs the destruction without calling delete() on the object. The second destructor, called the deleting destructor, calls delete() after destroying the object. Both destroy any virtual bases; a separate, non-virtual function, called the base object destructor, performs destruction of the object but not its virtual base subobjects, and does not call delete().

zeule: The written above is correct as far as I am aware. Since the destructor is mapped to to `Dispose()` and it makes no sense to call the non-deleting version from outside (if that `Dispose()` needs to be called at all), I again can't understand why the other one should be exposed.

tannergooding: ClangSharp is multiple parts, only one of which supports C#. It also is a general purpose wrapper over Clang itself, can expose the metadata for bindings via XML, and is used by the `win32metadata` project which itself is used for generating bindings for other languages, like Rust or even "modern C++" wrappers.

A power user may want to use this metadata to generate bindings for another language or even to synthesize their own types that "inherit" from a C++ type, in which case they may want to call the complete (non-deleting) destructor and call delete themselves.

zeule: Attempted to implement what you asked for. Take a look, please.

sources/libClangSharp/ClangSharp.cpp
```diff
@@ -4693,23 +4693,51 @@ CXCursor clangsharp_Cursor_getVBase(CXCursor C, unsigned i) {
     return clang_getNullCursor();
 }
 
-int64_t clangsharp_Cursor_getVtblIdx(CXCursor C) {
+namespace {
```
tannergooding, line 4696: nit: I'd likely put this at the top with the other helper functions (before `clangsharp_Cursor_getArgument`) and not put it into a namespace.

sources/libClangSharp/ClangSharp.h
```diff
@@ -722,6 +729,8 @@ CLANGSHARP_LINKAGE CXCursor clangsharp_Cursor_getUsedContext(CXCursor C);
 
 CLANGSHARP_LINKAGE CXCursor clangsharp_Cursor_getVBase(CXCursor C, unsigned i);
 
+CLANGSHARP_LINKAGE int64_t clangsharp_Cursor_getDtorVtblIdx(CXCursor C, CX_DestructorType dtor);
```
tannergooding, line 732: The corresponding entry needs to be added to just before https://github.com/microsoft/ClangSharp/blob/main/sources/ClangSharp.Interop/clangsharp.cs#L870 for the C# side, same with adding the enum definition for `CX_DestructorType`

sources/libClangSharp/ClangSharp.cpp
tannergooding: LGTM minus the two nits/fixes requested below.

sources/libClangSharp/ClangSharp.cpp
zeule: Are you OK with exporting all the symbols on non-MSVC platforms because they are not marked as hidden? All the helper functions (e.g. `isAttr()`) are currently exported. I can fix that, if you like.

tannergooding: There isn't any real harm to them being exported. If a power user wants to take a dependency on them, they can.

I'd also be fine with them being explicit and matching what MSVC exports.

zeule: There are no declarations for these functions in the header file, those who want to use them have to modify sources first. I'd say it is perfectly OK to hide them from the export table by default.

tannergooding (approved):

PR #244, opened 2021-06-25
Fix build with Clang 12 and install the native lib

### LINKED ISSUE #241: Does not build with Clang 12
The native project tries to link to clang libraries that do not exist (clangAST and clangFrontend). Clang 12 provides "clang-cpp" library instead.

zeule:
Closes #241.

### DIFF
--- a/sources/libClangSharp/CMakeLists.txt
+++ b/sources/libClangSharp/CMakeLists.txt
@@ -18,12 +18,6 @@ set(SOURCES
   CXType.h
 )
 
-set(LIBS
-  clangAST
-  clangFrontend
-  libclang
-)
-
 set(LLVM_SEARCH_PATHS
   ${PATH_TO_LLVM}
   ${PATH_TO_LLVM}/lib/cmake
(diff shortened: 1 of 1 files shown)


sources/libClangSharp/CMakeLists.txt
```diff
@@ -18,12 +18,6 @@ set(SOURCES
   CXType.h
 )
 
-set(LIBS
-  clangAST
-  clangFrontend
-  libclang
-)
```
tannergooding, line 25: Can you provide more details about how you built such that you didn't have `clangAST` or `clangFrontend`?

These exist in both the official release binaries (https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0, see `lib/libclangAST.a and friends) and in the binaries produced if built locally.

tannergooding: This CMake was also used to succesfully build libClangSharp for 12.0.0 on all supported platforms, so things were resolving correctly.

zeule: > Can you provide more details about how you built such that you didn't have clangAST or clangFrontend?

sys-devel/clang version 12.0.0 from Gentoo main tree. Neither it nor the clang package from Arch provide those libraries. You can view the file list for the Arch package [here](https://archlinux.org/packages/extra/x86_64/clang/). 

The clang package here installs no static (.a) archives at all.

zeule: [Gentoo configure](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-devel/clang/clang-12.0.0-r1.ebuild#n237)
[Arch build script](https://github.com/archlinux/svntogit-packages/blob/packages/clang/trunk/PKGBUILD)

tannergooding: This looks like its going to break windows as there is no `clang-cpp` produced.

I'm fine with taking it for Unix given it looks like the official binaries do create and ship `clang-cpp` by default.

zeule: Updated adding library check.

zeule:
Added `target_link_directories(ClangSharp PRIVATE ${LLVM_LIBRARY_DIRS})` to be on the sage side.

zeule:
> Added `target_link_directories(ClangSharp PRIVATE ${LLVM_LIBRARY_DIRS})` to be on the safe side.

No, that's unneeded, the import library path is set correctly by Clang's Targets.cmake.

sources/libClangSharp/CMakeLists.txt
```diff
@@ -37,12 +31,25 @@ find_package(Clang REQUIRED CONFIG
...
 endif()
+
+target_include_directories(ClangSharp PRIVATE ${CLANG_INCLUDE_DIRS})
+set_target_properties(ClangSharp PROPERTIES
+    PREFIX lib
+    VERSION ${PROJECT_VERSION}
+    SOVERSION ${PROJECT_VERSION}
+)
+
+include(GNUInstallDirs)
```
tannergooding, line 49: What's the behavior of this on Windows and is it going to do the "right" thing where Windows considers `*.dll` a `bin` file but `*.lib` a library file (in contrast to Unix which considers both `*.so` and `*.a` to be library files)?

zeule: On Windows _BINDIR is "bin", _LIBDIR is "lib", .dll files are `RUNTIME`, .lib files are `ARCHIVE'. CMake documents that `[here](https://cmake.org/cmake/help/latest/command/install.html#targets).

tannergooding (approved):

PR #249, opened 2021-07-10
Single quote must be escaped otherwise ''' is generated instead of '\''.

ronaldvanmanen:
Single quote must be escaped otherwise ''' is generated instead of '\\'' which results in a compiler error (e.g.: CS1011, C1012, etc.).

For example, the current version generates the following code for SDL_keycode.h:
<pre>
public enum SDL_KeyCode
{
    SDLK_QUOTE = (byte)(''')
}
</pre>
instead of 
<pre>
public enum SDL_KeyCode
{
    SDLK_QUOTE = (byte)('\'')
}
</pre>


tannergooding (approved):

tannergooding:
Thanks for the contribution!

PR #265, opened 2021-08-30
Add remaps for copy, move and deref operators

hermann-noll:
This PR adds remapped operator names for the copy, move and dereference operators. Previously this would have produced invalid C# code (e.g. `public unsafe void operator=(`. Unfortunately I could not find much guidance for "canonical C#" names, so "MoveFrom", "CopyFrom" and "Access" (for `->`) might not be the best choice.

Note that I did not use return types for the tests as this seems to be its own bug with returning references (e.g. C++ `MyStruct& abc() { return *this; }` would produce C# `MyStruct* abc() { return *this; }`.


tannergooding:

PR #266, opened 2021-08-30
Fix imported constructors and instance methods

hermann-noll:
Constructors were supported when they had bodies and were generated as actual C# constructors, but trying to import them produced invalid C#, e.g:

```csharp
    {
        [DllImport("", CallingConvention = CallingConvention.ThisCall, EntryPoint = "...", ExactSpelling = true)]
        public static extern MyStruct();
    }
```

So three things had to be fixed:

  - The name has to be remapped to not conflict with C# constructors (static or not) <br/> Also copy/move constructors will conflict with each other thus they get different names
  - A return type has to be added (Clang said `void` although it *might* also be a pointer to the struct?)
  - A `pThis` parameter has to be added. <br/> This point applied to all imported instance methods.

Here is an example of the result of this PR:

```csharp
    {
        [DllImport("", CallingConvention = CallingConvention.ThisCall, EntryPoint = "...", ExactSpelling = true)]
        public static extern unsafe void Constructor(MyStruct* pThis);

        [DllImport("", CallingConvention = CallingConvention.ThisCall, EntryPoint = "...", ExactSpelling = true)]
        public static extern unsafe void CopyConstructor(MyStruct* pThis, [NativeTypeName("const MyStruct &")] MyStruct* other);

        [DllImport("", CallingConvention = CallingConvention.ThisCall, EntryPoint = "...", ExactSpelling=true)]
        public static extern unsafe void MoveConstructor(MyStruct* pThis, [NativeTypeName("MyStruct &&")] MyStruct* other);

        [DllImport("", CallingConvention = CallingConvention.ThisCall, EntryPoint = "...", ExactSpelling = true)]
        public static extern unsafe void MyMethod(MyStruct* pThis);
    }
```

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/Abstractions/FunctionOrDelegateDesc.cs
+++ b/sources/ClangSharp.PInvokeGenerator/Abstractions/FunctionOrDelegateDesc.cs
@@ -214,6 +214,14 @@ namespace ClangSharp.Abstractions
                 : Flags & ~FunctionOrDelegateFlags.IsCxxConstructor;
         }
 
+        public bool NeedsThisParameter
+        {
+            get => (Flags & FunctionOrDelegateFlags.NeedsThisParameter) != 0;
+            set => Flags = value
+                ? Flags | FunctionOrDelegateFlags.NeedsThisParameter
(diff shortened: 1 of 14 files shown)


tannergooding:

PR #267, opened 2021-08-30
Replace generation of protected entries

hermann-noll:
Protected C++ entries are also generated as protected C# entries. This is afaik always wrong as protected is not allowed in neither namespaces nor structs. This PR thus replaces this access modifier with `public` as my reasoning was that using applications might want to access those to e.g. implement new inheriting classes. But this could very well also be a config flag.
I deliberately included a virtual protected method with explicit vtables and without function pointers to test delegates and vtable fields.

During this feature there was an issue with delegates not being generated when the implementation is not user-provided (e.g. `virtual ~MyStruct() = default;`)

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -410,9 +410,12 @@ namespace ClangSharp
 
         private void VisitFunctionDecl(FunctionDecl functionDecl)
         {
-            if (!functionDecl.IsUserProvided)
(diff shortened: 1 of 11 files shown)


tannergooding:
> This is afaik always wrong as protected is not allowed in neither namespaces nor structs. This PR thus replaces this access modifier with public as my reasoning was that using applications might want to access those to e.g. implement new inheriting classes. But this could very well also be a config flag.

I've been tending to use `internal` as an alternative default here. This is a concept where C++ does not have a clear mapping to .NET and where its likely needs manual intervention from the user no matter what ClangSharp generates.

tannergooding:
Closing as there has been no response to the feedback. Please feel free to re-open and address them.

PR #268, opened 2021-08-31
Fix std::nullptr_t

hermann-noll:
The usual way of using the nullptr type of C++ is by using the STL provided `std::nullptr_t` which translates into `typedef decltype(nullptr) nullptr_t;`. Previously this would have produced a diagnostic from ClangSharp and `decltype(nullptr)` would be used in the output. 
Although this will surely not work for every possible of usage of `decltype` I have added a simple forwarding to the underlying type as reported by Clang which will then map to the built-in `CXType_NullPtr` type. But this would have produced declarations like `null myVariable` which is incorrect thus I have changed this to `IntPtr`. 

Alternatively we could output (or relay to the user) an additional type which would better represent the C++ nature of `std::nullptr_t` like the following:

```csharp
public readonly struct Nullptr
{
	private readonly IntPtr @value;
	public static readonly Nullptr Null = default;
}
					
public class Program
{
	static extern void A(Nullptr p);
	
	public static void Main()
	{
		A(new Nullptr());
		A(Nullptr.Null);
		A(default);
	}
}
```

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -1541,7 +1541,7 @@ namespace ClangSharp
 
                     case CXTypeKind.CXType_NullPtr:
                     {
-                        name = "null";
+                        name = "IntPtr";
                         break;
                     }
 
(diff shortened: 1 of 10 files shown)


hermann-noll:
Regarding the ubuntu CI errors, this seems more like a missing installation on the machine?

tannergooding:
I'm not convinced returning `IntPtr` is the right choice here either, `void*` is probably the closest to the right choice, but its a concept where there is no real mapping between C/C++ and .NET and so its a concept that might never be able to have good/correct codegen here particularly given its use-case simply as a way to provide "overload resolution" when `null` is passed in and at least two other overloads accept pointer types in the same position.

> Alternatively we could output (or relay to the user) an additional type which would better represent the C++ nature of std::nullptr_t

This is not ABI compatible with C/C++.

PR #272, opened 2021-09-29
Updating Cursor and Type to avoid repeated hot path allocations

### LINKED ISSUE #260: Performance of Cursor's CursorChildren (CXCursor.VisitChildren)
When profiling our application that uses ClangSharp, I've noticed that the `Cursor` class's `CursorChildren` lazy property evaluation takes a significant amount of time on profiles, specifically inside of the `CXCursor.VisitChildren` call.

It looks like 75% of that time is being spent inside of `MarshalNative::GetFunctionPointerForDelegateInternal` (of which nearly all of the time is inside of `COMDelegate::ConvertToCallback`).

Given the fact that this is likely to be a hotspot across many applications, is there an opportunity to avoid such a penalty by avoiding using the more "generic" `CXCursor.VisitChildren` in lieu of a custom native method that is P/Invoked to populate a `List` of the immediate children of a given `Cursor` using a native function (therefore avoiding interop of the otherwise-uninteresting delegate)?

If that's feasible, am I correct in my understanding of the project hierarchy here that such code would live within the libClangSharp project?

Thanks,

Daniel Jennings
Valve

tannergooding:
This resolves #260

### DIFF
--- a/sources/ClangSharp/Cursors/Cursor.cs
+++ b/sources/ClangSharp/Cursors/Cursor.cs
@@ -3,6 +3,8 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using ClangSharp.Interop;
 
 namespace ClangSharp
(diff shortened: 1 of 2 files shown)


danieljennings:
Thanks! Those changes look great; they look like a smarter version of what I was doing locally to workaround this (including the .NET 5+ stuff that I hadn't seen used before today).

Thanks for taking the time to do this work, and for working on this project in general.

PR #288, opened 2021-11-11
Add support for a new --with-namespace option

tannergooding:


ankyo:
how to use this option? does it has a example?

PR #329, opened 2022-03-28
Update link to win32metadata.

teo-tsirpanis:
It was moved from the dotnet org to microsoft.


tannergooding (approved):
I actually think it was always under microsoft and I just copy/paste errored 😅

PR #340, opened 2022-04-20
Fix linking to Clang libraries on Linux

### LINKED ISSUE #336: Does not build on Linux
On Gentoo the build fails with:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lclangAST
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lclangFrontend
```
Both libraries do not exist in my system.

zeule:
Linux distributions prefer dynamic libraries but we'd like to link to
Clang statically. Therefore we return the test for clang targets but
extend it to prefer static Clang libraries and fall back to the
clang-cpp library when the static ones are unavailable.

Fixes #336.

### DIFF
--- a/sources/libClangSharp/CMakeLists.txt
+++ b/sources/libClangSharp/CMakeLists.txt
@@ -33,13 +33,20 @@ set(LLVM_SEARCH_PATHS
   ${PATH_TO_LLVM}/share/llvm/cmake/
 )
 
+option(LLVM_USE_STATIC "Link to static versions of LLVM/Clang libraries" ON)
+
 find_package(Clang REQUIRED CONFIG
              PATHS ${LLVM_SEARCH_PATHS}
              NO_DEFAULT_PATH)
 
 add_library(ClangSharp SHARED ${SOURCES})
 
-target_link_libraries(ClangSharp PRIVATE clangAST clangFrontend libclang)
+if (LLVM_USE_STATIC)
+    target_link_libraries(ClangSharp PRIVATE clangAST clangFrontend)
+else()
+    target_link_libraries(ClangSharp PRIVATE clang-cpp)
+endif()
+target_link_libraries(ClangSharp PRIVATE libclang)
 
 target_include_directories(ClangSharp PRIVATE ${CLANG_INCLUDE_DIRS})
 


tannergooding:
I don't think this is a change that we want to take.

It is not unreasonable to require devs to build a local copy of LLVM before building ClangSharp and that will ensure that it a consistent experience is given everywhere.

zeule:
And how am I supposed to install that custom build in PATH where I already have another LLVM installed?

tannergooding:
> And how am I supposed to install that custom build in PATH where I already have another LLVM installed?

Why would you need to install it? `llvmorg-14.0.0` built locally should be (ignoring any signing or anything) identical in contents to what is officially produced.

For the purposes of building `libClangSharp`, you simply follow the steps here: https://github.com/dotnet/ClangSharp#building-native
```
# assuming from some ~/repos folder
git clone --single-branch --branch llvmorg-14.0.0 https://github.com/llvm/llvm-project
cd llvm-project
mkdir artifacts/bin -p
cd artifacts/bin
cmake -DLLVM_ENABLE_PROJECTS=clang -DCMAKE_INSTALL_PREFIX=../install -G "Unix Makefiles" ../../llvm
make install
```

After its built, you:
```
# assuming from some ~/repos folder
git clone https://github.com/dotnet/clangsharp
cd clangsharp
mkdir artifacts/bin -p
cd artifacts/bin
cmake -DPATH_TO_LLVM=/home/username/repos/llvm/artifacts/install -DCMAKE_INSTALL_PREFIX=../install -G "Unix Makefiles" ../..
make install
```

zeule:
Thank you, of course, but I refer to another problem. The problem is common for POSIX systems and it is exactly the one why we avoid installing software without the system package manager.

ibClangSharp.so links to the libclang.so.xx, which is already installed in my (and any other Linux) system. Now I compile Clang myself and I want ClangSharpPInvokeGenerator executable to be in PATH and it needs to find the libclang.so.xx, which thus needs to be in ld.so.conf (i.e. in one of the system directories for dynamic libraries). If I put it in a directory, searched before the one with the system libclang.so.xx, I ensure ClangSharpPInvokeGenerator links at runtime with the library it was built, but all the other system tools now links to my custom library. And all of them are potentially broken now. To minimise the risk to break my system, I have to ensure my custom libclang.so.xx is built with the same options as the system clang package, and if I use LTO (I do), even with the same version of the compiler. That's too much of a burden and nobody wants to do that. My only option is to install clangsharp with my custom clang into its own prefix and provide a startup script that will instruct dynamic linker to search this custom prefix for the custom ibclang.so.xx. But I'm fed up with 10 copies of webkit installed already with every fucking Electron application, and I'm angry; the idea of having multiple copies of LLVM is a stressor which triggers me instantly and I better re-design all the scripting in my project to switch over from C# to Python or even Lua than to participate in this hell of creating senseless copies of big packages.

At this stage I realise it is much simpler to teach the clangsharp build script to link to the system clang libraries, which is done here...

If you are afraid it will not work automatically, let's put it under a CMake `option()` ?

tannergooding:
That's why `libClang` and `libClangSharp` are distributed with `ClangSharpPInvokeGenerator` and are provided via ` NuGet` package for other .NET tools to achieve the same.

Because .NET will resolve and load them from the same directory as the .NET executable. `ClangSharp.Interop` additionally has an extra step (https://github.com/dotnet/ClangSharp/blob/main/sources/ClangSharp.Interop/Extensions/clang.ResolveLibrary.cs) that will try to resolve the relevant alternative names/etc so it works on more systems and platforms.

zeule:
Polluting my system with god only knows where and by whom produced binaries from nuget (not only with respect to this project, but in general) is out of the question, of course (remember, Linux systems do not have anti-virus software). I can tolerate them to a certain degree in a sandbox where packages are built, but not inside a real filesystem. No, thanks, no way.

tannergooding:
The TL;DR: ClangSharp already supports several distro's and users are able to successfully use the tooling and libraries there. Needing to support several different distro's is already significantly more complex than what's required for Windows or MacOS.

I have little interest in opening the doors to specializing support further for additional distro's, especially when those distro's will work with the current setup and its already the norm that various binaries, including both managed and native binaries, will ship SxS with a given .NET application.

> Polluting my system with god only knows where and by whom produced binaries from nuget (not only with respect to this project, but in general) is out of the question

That's ultimately a decision for you to make. If you install the .NET runtime, then you'll only get the "in box" bits globally installed. Other 3rd party dependencies or dependencies that aren't "in box" ship SxS with the .NET app (including both managed and native dependencies). This is the norm.

If you want to go with portable .NET apps, then they are generally "self contained" and even the runtime ships SxS with the app rather than being globally installed. This is the preferred approach by many tool authors so that getting the app is simple and can be "xcopyable" (put it on a flash drive and it works on any machine you take it to).

> remember, Linux systems do not have anti-virus software

There are plenty of distro's that do support anti-virus and many offerings are available. As Linux becomes more prevalent, it's more likely to be targeted by malicious users and needing anti-virus will become the norm. Simply running as non-sudo is also not enough given various hardware exploits that arise from time to time.

zeule:
Well, all distros (there was no need to check, see above, but I checked Fedora, Debian/Ubuntu, and Arch) provide clang-spp.so. I'm pretty sure all other do package clang with this dynamic library too. I understand your problem comes from the desire to distribute the nuget package and thus you need static linking because you don't know what clang version is installed at the target machine. This is a flawed way to distribute software for Linux, but you probably already know that.

> That's ultimately a decision for you to make. 

I would be happy to make a decision, but I can't because my decision is to package this software and install it globally.

> If you want to go with portable .NET apps…

No, I don't. And yes, I'm aware of the way software is distributed in the Windows world. And that's one of the main reasons I don't use Windows other than to host Steam client.

> As Linux becomes more prevalent, it's more likely to be targeted by malicious users and needing anti-virus will become the norm.

If desktop Linux evolves into a popular desktop platform, so that malicious software becomes evident in the wild, I'm pretty sure there will be another platform available by that time, free of such design choices targeted at making it popular for general public no matter what.

zeule:
Replaced tests with explicit build option.

tannergooding:
> This is a flawed way to distribute software for Linux, but you probably already know that.

It is not a flawed way to distribute .NET applications, it is the expected norm.

This is not something that this project is willing to take on and support. If you would like to support this yourself, I recommend you fork the project and maintain the changes independently.

ClangSharp is going to continue targeting the .NET norm and the industry standard practices that help ensure software can be easily versioned, distributed, and maintained over time.

kant2002:
@tannergooding can you reconsider closing this PR? If you take source based Linux distribution into context, like Gentoo, the requirement to avoid Nuget does not seems `exotic` anymore. It maybe unusual to you, but this is norm for that distro. Ocaml, Java and other ecosystems who has their own package manager, also has packages which locally built.

PR #341, opened 2022-04-21
Allow private access specifiers on functions

saucecontrol:
`private` access specifier on a function identifier is getting overridden to `internal` by the check meant to prevent `private` on non-nested types.

The XML generator is already doing the right thing.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -491,7 +491,7 @@ namespace ClangSharp
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -3677,7 +3677,7 @@ private bool IsPrimitiveValue(Type type)
             {
                 return IsPrimitiveValue(autoType.CanonicalType);
             }
-            else if (type is BuiltinType builtinType)
```
saucecontrol, line 3680: this was giving a compiler warning for unused variable

tannergooding (approved):

PR #344, opened 2022-04-30
Add support for varargs.

Berrysoft:
According to microsoft/win32metadata#917 and https://github.com/microsoft/win32metadata/issues/33 , the win32 metadata doesn't recognize the varargs declarations, which gives much inconvenience. This PR adds simple dll-import support for varargs, and generate C# code with undocumented keyword `__arglist`.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -1922,6 +1922,14 @@ namespace ClangSharp
 
         private CallingConvention GetCallingConvention(Cursor cursor, Cursor context, Type type)
         {
+            if (cursor is FunctionDecl functionDecl)
+            {
+                if (functionDecl.IsVariadic)
+                {
+                    return CallingConvention.Cdecl;
+                }
+            }
+
             if (cursor is NamedDecl namedDecl)
             {
                 if (TryGetRemappedValue(namedDecl, _config.WithCallConvs, out var callConv, matchStar: true))
--- a/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleWindows/FunctionDeclarationDllImportTest.cs
+++ b/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleWindows/FunctionDeclarationDllImportTest.cs
@@ -407,7 +407,7 @@ namespace ClangSharp.Test
     public static partial class Methods
     {
         [DllImport(""ClangSharpPInvokeGenerator"", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
-        public static extern void MyFunction(int value, __arglist );
+        public static extern void MyFunction(int value, __arglist);
     }
 }
 ";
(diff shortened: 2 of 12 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -1922,6 +1922,14 @@ private Type[] GetBitfieldCount(RecordDecl recordDecl)
 
         private CallingConvention GetCallingConvention(Cursor cursor, Cursor context, Type type)
         {
+            if (cursor is FunctionDecl functionDecl)
+            {
+                if (functionDecl.IsVariadic)
+                {
+                    return CallingConvention.Cdecl;
+                }
+            }
```
tannergooding, line 1931: This needs to support remapping still. The default should definitely be `Cdecl`, but ClangSharp supports users doing any kind of remapping, even ones that technically may not make sense.

tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleWindows/FunctionDeclarationDllImportTest.cs
```diff
@@ -395,5 +395,24 @@ public static partial class Methods
...
+            const string InputContents = @"extern ""C"" void MyFunction(int value, ...);";
+
+            const string ExpectedOutputContents = @"using System.Runtime.InteropServices;
+
+namespace ClangSharp.Test
+{
+    public static partial class Methods
+    {
+        [DllImport(""ClangSharpPInvokeGenerator"", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        public static extern void MyFunction(int value, __arglist );
```
tannergooding, line 410: It would be good to specialcase `Type == "__arglist"` so we can avoid the stray space at the end.

tannergooding:
Sorry for the delay in reviewing this. It generally looks correct to me but has a couple fixups that would be good to include.

There should ideally be some handling for a variadic function pointer as well. C# doesn't support them, but ClangSharp still needs to generate something (likely just a diagnostic indicating it couldn't be handled and falling back to `void*`).

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
Berrysoft: The document of [`__stdcall`](https://docs.microsoft.com/en-us/cpp/cpp/stdcall) indicates that even the vararg function is marked as `__stdcall`, the compiler will makes it a `__cdecl`. Therefore I force `Cdecl` here.

tannergooding: Yes, but that's not necessarily true for any and all potential platforms. ClangSharp has reasonable defaults and allows devs to override the default to fit their own needs.

Berrysoft: I think we are generating proper dll imports. If a programmer writes a method like
``` cpp
extern "C" int __stdcall foo(...);
```
The compiler won't raise errors. However, if we generate bindings like
``` csharp
[DllImport("foo.dll", CallingConvention = CallingConvention.StdCall)]
static extern int foo(__arglist);
```
The program will raise runtime `System.TypeLoadException` says that varargs function should use cdecl.

PR #358, opened 2022-07-11
Do not remove parethensis

### LINKED ISSUE #357: Macros did not apply correctly.
original C++ file
```cpp
typedef struct
{
	unsigned char *buf;
	int size;
} context_t;

#define PCONT ((context_t *)pcontext)

int buf_close(void *pcontext)
{
	PCONT->buf=0;
	return 0;
}
```

produced C#
```csharp
namespace test
{
    public unsafe partial struct context_t
    {
        [NativeTypeName("unsigned char *")]
        public byte* buf;

        public int size;
    }

    public static unsafe partial class Methods
    {
        public static int buf_close(void* pcontext)
        {
            (context_t*)(pcontext)->buf = null; /// Issue is here !!!
            return 0;
        }
    }
}
```

kant2002:
That's potentially risky move on real codebases, but that's not captured anywhere in tests, so I will move ahead with a change.

Would be good if somebody test that on some larger codebase and check how it works for them.

Most other changes is only in testing infra. I take liberty and do following change operating from following premises.
- there only 2 radically different output formats: CSharp and Xml
- Most codegeneration would be platform neutral.
- You will need to override tests based on platform needs (like it is right now)

I do trying not create sets of identical tests for each platform, since it's a bit boring and timeconsuming to validate. For example using existing approach not clear what's platform-independent behaviour and what is OS specific behavior of ClangSharp. After this change all platform independent tests may sit in Base folder, and additional tests and/or augmentations will be in same location as it is now.

Fixes #357

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
@@ -1612,7 +1612,7 @@ namespace ClangSharp
 
             if (!memberExpr.IsImplicitAccess || isForDerivedType)
             {
-                var memberExprBase = memberExpr.Base.IgnoreParens.IgnoreImplicit;
+                var memberExprBase = memberExpr.Base.IgnoreImplicit;
 
                 if (isForDerivedType)
                 {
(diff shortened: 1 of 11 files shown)


tannergooding:
Validated against a few various projects and no issues seen.

PR #368, opened 2022-08-12
Version check skip option and patch verison leniency

Beyley:


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -21,8 +21,8 @@ public sealed partial class PInvokeGenerator : IDisposable
         private static readonly Encoding s_defaultStreamWriterEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
         private static readonly Regex s_needsSystemSupportRegex = new Regex(@"\b(?:Guid|IntPtr|UIntPtr)\b", RegexOptions.Compiled);
 
-        private const string ExpectedClangVersion = "14.0.0";
-        private const string ExpectedClangSharpVersion = "14.0.0";
+        private const string ExpectedClangVersion = "version 14.0";
+        private const string ExpectedClangSharpVersion = "version 14.0.";
```
tannergooding, line 25: nit: Should be fine to keep these inline:
```suggestion
        private const string ExpectedClangVersion = "version 14.0";
        private const string ExpectedClangSharpVersion = "version 14.0";
```

tannergooding (approved):

Issue #339, opened 2022-04-20
generate-helper-types with single-file is broken


mgood7123:
```
PS D:\IMPORTANT\source\repos\WindowsProject1\AndroidUI.Native> ls


    Directory: D:\IMPORTANT\source\repos\WindowsProject1\AndroidUI.Native


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         4/20/2022   8:12 PM           1518 AndroidUI.Native.vcxitems
-a----         4/19/2022   5:09 PM           1378 TmpPtr.h
-a----         4/20/2022   8:14 PM          47296 SkXn.h
-a----         4/20/2022  10:20 PM             39 a.cs
-a----         4/18/2022   5:24 PM          21738 SkTypes.h
-a----         4/20/2022   7:09 PM           3392 SkNx_IMPL.cpp
-a----         4/14/2022   1:33 PM           1521 SkSafe_math.h
-a----         4/14/2022   1:43 PM           2700 SkFloatBits.h
-a----         4/14/2022   1:44 PM           6877 SkScalar.h
-a----         4/20/2022   8:12 PM            885 AndroidUI.Native.vcxitems.filters
-a----         4/14/2022   1:40 PM          10111 SkFloatingPoint.h
-a----         4/17/2022   2:19 PM          30147 SkNx_sse.h
-a----         4/17/2022   2:22 PM          23245 SkNx_neon.h


PS D:\IMPORTANT\source\repos\WindowsProject1\AndroidUI.Native> ClangSharpPInvokeGenerator.exe -f .\SkXn.h -l "runtimes/win-x64/native/AndroidUI.Native.Windows.dll"           -n Bindings -m Windows -c compatible-codegen -c generate-helper-types -o I_DO_NOT_EXIST.cs
Processing '.\SkXn.h'
Diagnostics for binding generation of .\SkXn.h:
    Error (Line 35, Column 11 in ./SkXn.h): Unsupported anonymous named declaration: 'Namespace'.
    Error (Line 173, Column 12 in ./SkXn.h): Unsupported declaration: 'ClassTemplatePartialSpecialization'. Generated bindings may be incomplete.
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
[...]

PR #391, opened 2022-09-18
Ensure generate-helper-types works with single-file mode

tannergooding:
This resolves https://github.com/dotnet/ClangSharp/issues/339


zh4ui:
Hi, it seems generate-helper-types does not work in multi-file mode? I tried with 15.0.2 and received error like "cannot open stream to write".

PR #405, opened 2022-10-28
Add support for --with-packing

DaZombieKiller:
Allows a user to specify an exact value for `StructLayoutAttribute.Pack`, which is useful for controlling packing with a constant value defined in C#.

For example, `--with-packing ExampleStruct=InteropDetails.PackConstant` given the following header:
```cpp
struct ExampleStruct
{
};
```
Will output the following C# definition of `ExampleStruct`:
```cs
[StructLayout(LayoutKind.Sequential, Pack = InteropDetails.PackConstant)]
public partial struct ExampleStruct
{
}
```

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -746,15 +746,10 @@ internal partial class CSharpOutputBuilder : IOutputBuilder
             WriteIndented("[StructLayout(LayoutKind.");
             Write(desc.LayoutAttribute.Value);
 
-            if (desc.Layout.PackOverride is { } packOverride)
+            if (desc.Layout.Pack != null)
             {
                 Write(", Pack = ");
-                Write(packOverride);
-            }
-            else if (desc.LayoutAttribute.Pack != 0)
-            {
-                Write(", Pack = ");
-                Write(desc.LayoutAttribute.Pack);
+                Write(desc.Layout.Pack);
             }
 
             WriteLine(")]");
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -740,18 +740,18 @@ internal partial class CSharpOutputBuilder : IOutputBuilder
             WriteLine("\"]/*' />");
         }
 
-        if (desc.LayoutAttribute is not null || desc.Layout.PackOverride is not null)
+        if (desc.LayoutAttribute is not null)
         {
             AddUsingDirective("System.Runtime.InteropServices");
             WriteIndented("[StructLayout(LayoutKind.");
-            Write(desc.LayoutAttribute?.Value ?? LayoutKind.Sequential);
+            Write(desc.LayoutAttribute.Value);
 
             if (desc.Layout.PackOverride is { } packOverride)
             {
                 Write(", Pack = ");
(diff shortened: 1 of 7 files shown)


tests/ClangSharp.PInvokeGenerator.UnitTests/PInvokeGeneratorTest.cs
```diff
@@ -106,6 +106,7 @@ private static async Task ValidateGeneratedBindingsAsync(string inputContents, s
             WithTransparentStructs = withTransparentStructs,
             WithTypes = withTypes,
             WithUsings = withUsings,
+            WithPackOverrides = null,
```
tannergooding, line 109: I'd probably just call this `WithPacking`

sources/ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -1288,4 +1294,15 @@ private static Option<string[]> GetWithUsingOption()
             AllowMultipleArgumentsPerToken = true
         };
     }
+
+    private static Option<string[]> GetWithPackOverrideOption()
+    {
+        return new Option<string[]>(
+            aliases: new string[] { "--with-pack-override", "-wpo" },
```
tannergooding, line 1301: and for this one probably just `--with-packing`

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -1479,6 +1479,7 @@ private void VisitRecordDecl(RecordDecl recordDecl)
                     Size32 = size32,
                     Size64 = size64,
                     Pack = alignment < maxAlignm ? alignment : 0,
```
tannergooding, line 1481: Why not just pass the value into `Pack` and have it fallback to `((alignment < maxAlignm) ? alignment : 0)`?

sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
```diff
@@ -740,13 +740,18 @@ public void BeginStruct(in StructDesc desc)
             WriteLine("\"]/*' />");
         }
 
-        if (desc.LayoutAttribute is not null)
+        if (desc.LayoutAttribute is not null || desc.Layout.PackOverride is not null)
```
tannergooding, line 743: Why not just have `LayoutAttribute` take this into account instead?

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
DaZombieKiller: `PackOverride` is a string, so that it can be set to the name of a constant. If I used `Pack` it'd need to have its type changed to accommodate that.

tannergooding: Right, that seems fine overall. Just use `null` or `empty string` to denote `0` instead.

sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
DaZombieKiller: That would definitely make more sense, addressed that in my latest commit.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
DaZombieKiller: Addressed in most recent commit.

tannergooding (approved):

PR #431, opened 2023-03-05
Workaround some bugs in a few boolean returning interop methods

### LINKED ISSUE #416: libClangSharp - some bool flags inverted
There are a number of cursor properties which when accessed through libClangSharp return the inverse of what they should:

```
clangsharp_Cursor_getIsCopyOrMoveConstructor
clangsharp_Cursor_getIsDeleted
clangsharp_Cursor_getIsExplicitlyDefaulted
```

There are two more that I didn't test - but from looking at the source code look like they may also be inverted:

```
clangsharp_Cursor_getIsInheritingConstructor
clangsharp_Cursor_getIsDelegatingConstructor
```

As a test case; when inspecting the CXCursor of the constructors in the following example, the inspected properties don't match the actual signature. I've put a comment above each constructor  showing what libClangSharp reports.

```
class SomeClass;
class MyTest {
 public:
  // libClangSharp: IsCopyOrMove = true, IsCopy = false, IsMove = false, IsDeleted = true, IsExplicitlyDefaulted = true
  explicit MyTest(std::shared_ptr<SomeClass> someClassInstance, float someFloat = {});

  // libClangSharp: IsCopyOrMove = false, IsCopy = true, IsMove = false, IsDeleted = false, IsExplicitlyDefaulted = true
  MyTest(const MyTest& other) = delete;

  // libClangSharp: IsCopyOrMove = false, IsCopy = false, IsMove = true, IsDeleted = true, IsExplicitlyDefaulted = false
  MyTest(MyTest&& other) = default;
}
```

tannergooding:
This resolves #416

### DIFF
--- a/sources/ClangSharp.Interop/Extensions/CXCursor.cs
+++ b/sources/ClangSharp.Interop/Extensions/CXCursor.cs
@@ -906,7 +906,7 @@ public unsafe partial struct CXCursor : IEquatable<CXCursor>
 
     public bool IsConversionFromLambda => clangsharp.Cursor_getIsConversionFromLambda(this) != 0;
 
-    public bool IsCopyOrMoveConstructor => clangsharp.Cursor_getIsCopyOrMoveConstructor(this) != 0;
+    public bool IsCopyOrMoveConstructor => clangsharp.Cursor_getIsCopyOrMoveConstructor(this) == 0;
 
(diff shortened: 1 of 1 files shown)


tannergooding:
@tdmowrer, its a temporary fix to ensure the managed side is correct.

I'll be fixing the native side and getting new native packages up with v16.0.0 when that releases in the next month or so (building all native libraries for all supported targets is "very" expensive and takes several days).

PR #440, opened 2023-04-08
Fix generate-macro-bindings generation

### LINKED ISSUE #439: generate-macro-bindings only works when generating bindings from C++ files
Hello, I have been trying to use this project to generate bindings from the XCB library. However, I noticed that when generating bindings from C headers, no constants are generated from the macros.

I have dived into the code and I have found that the generator generates auto constants from the macros in a first pass and then translates the entire ast into C# in a second pass.

Currently, that only works for C++ files.

GasInfinity:
Fixes macro bindings generation by changing 'auto' to __auto_type. Fixes #439


GasInfinity:
@dotnet-policy-service agree

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitPreprocessedEntity.cs
```diff
@@ -34,7 +34,7 @@ private unsafe void VisitMacroDefinitionRecord(MacroDefinitionRecord macroDefini
             var macroName = $"ClangSharpMacro_{macroDefinitionRecord.Spelling}";
 
             _ = _fileContentsBuilder.Append('\n');
-            _ = _fileContentsBuilder.Append($"const auto {macroName} = ");
+            _ = _fileContentsBuilder.Append($"const __auto_type {macroName} = ");
```
tannergooding, line 37: This overall LGTM, but there should be an explicit test added to https://github.com/dotnet/ClangSharp/blob/main/tests/ClangSharp.PInvokeGenerator.UnitTests/CTest.cs ensuring macro handling "works".

You can probably just copy the `MacroTest` from: https://github.com/dotnet/ClangSharp/blob/main/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpLatestWindows/VarDeclarationTest.cs#L72-L91 and ensure the `C` specific command line args are passed in.

It would be likewise be nice if we could continue using `auto` for C++ and only use `__auto_type` for C. It shouldn't make a difference, but not relying on compiler specific extensions can help avoid weird edge cases and make things more standardized overall. You should be able to determine which to use by checking the `macroDefinitionRecord.Handle.Language`

GasInfinity: The problem with doing that check is that I should also check the C++ standard used (preC++11 code would not compile if we don't do that). I couldn't find any method in clang-c that gives me the language features of the cursor, and I also see that we don't save the used standard inside `PInvokeGeneratorConfiguration`.

tannergooding: Please feel free to pass through the standard to the `PInvokeGeneratorConfiguration` if that helps determine the version to use.

GasInfinity:
If you want, I could squash the commits that only fixes some silly errors to keep the commit history clean

tannergooding (dismissed):

tests/ClangSharp.PInvokeGenerator.UnitTests/CTest.cs
```diff
@@ -138,6 +138,28 @@ public unsafe partial struct _MyOtherUnion
...
+{{
+    public static partial class Methods
+    {{
+        [NativeTypeName(""#define MyMacro1 5"")]
+        public const int MyMacro1 = 5;
+
+        [NativeTypeName(""#define MyMacro2 MyMacro1"")]
+        public const int MyMacro2 = 5;
+    }}
+}}
```
tannergooding, line 160: Since you aren't doing interpolation:
```suggestion
        var expectedOutputContents = @"namespace ClangSharp.Test
{
    public static partial class Methods
    {
        [NativeTypeName(""#define MyMacro1 5"")]
        public const int MyMacro1 = 5;

        [NativeTypeName(""#define MyMacro2 MyMacro1"")]
        public const int MyMacro2 = 5;
    }
}
```

GasInfinity: Yes, sorry, I missed that

tannergooding (approved):

tannergooding:
Thanks!

Issue #415, opened 2023-01-23
Read behaviour for signed bitfields is inconsistent with Clang, MSVC and GCC


DaZombieKiller:
When generating an interop structure for the following type:
```cpp
struct PackedVertex
{
    union
    {
        int i;
        struct
        {
            int x : 11;
            int y : 11;
            int z : 10;
        };
    };
};
```
`ClangSharpPInvokeGenerator` produces accessors like the following:
```cs
[NativeTypeName("int : 11")]
public int x
{
    get
    {
        return _bitfield & 0x7FF;
    }

    set
    {
        _bitfield = (_bitfield & ~0x7FF) | (value & 0x7FF);
    }
}
```
Which does not appear to handle the sign of each component properly.

[...]

PR #464, opened 2023-07-08
Add support for InlineArray, generate readonly instance members, and fix signed bitfield handling

tannergooding:
This resolves https://github.com/dotnet/ClangSharp/issues/415


kkukshtel:
Should we start seeing `inlinearray` on generated output if we switch to dotnet 8?

PR #471, opened 2023-09-18
Allow equals sign in value for parsed KVP Pairs

kkukshtel:
Parses KVP arg into a substring with only ever two parts based around the boundary of the first equals sign. Subsequent equals signs are ignored and and make their way into the final parsed value as is.

### DIFF
--- a/sources/ClangSharpPInvokeGenerator/Program.cs
+++ b/sources/ClangSharpPInvokeGenerator/Program.cs
@@ -852,9 +852,9 @@ public class Program
 
         foreach (var keyValuePair in keyValuePairs)
         {
-            var parts = keyValuePair.Split('=');
+            var parts = keyValuePair.Split('=',2);
 
-            if (parts.Length != 2)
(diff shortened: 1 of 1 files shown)


sources/ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -852,9 +852,9 @@ private static void ParseKeyValuePairs(IEnumerable<string> keyValuePairs, List<s
 
         foreach (var keyValuePair in keyValuePairs)
         {
-            var parts = keyValuePair.Split('=');
+            var parts = keyValuePair.Split('=',2);
```
tannergooding, line 855:
```suggestion
            var parts = keyValuePair.Split('=', 2);
```

tannergooding (approved):

tannergooding:
Failures look to be from the new .NET 8 preview SDK and are unrelated.

Going to take a look and get those resolved so I can merge this.

PR #521, opened 2024-01-15
Transparent fnptr structs

TechPizzaDev:


tannergooding:
@TechPizzaDev, this has been stuck in draft for a while. I think we'd need tests, a clearer use-case, and some consideration for the pitfalls like fnptrs not being safe to compare or equate.

PR #525, opened 2024-02-02
Fix native crashes / asserts in libClangSharp and CXXRecordDecl Destructor property.

ceresgalax:
* Work around a clang crash in clangsharp_Cursor_getLambdaStaticInvoker when CRD->getLambdaCallOperator() returns null and CRD->getLambdaStaticInvoker() is called.

* clangsharp_Cursor_getNumAttrs: Calling D->getAttrs() will assert if D->hasAttrs() returns false. Fix by checking D->hasAttrs() first.

* A CXXRecordDecl may return null for getDestructor. Change the C# CXXRecordDecl Destructor property to be nullable to support this case.

Hopefully these changes make inspecting ClangSharp objects in the debugger much less likely to crash.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -717,7 +717,7 @@ public partial class PInvokeGenerator
                 {
                     outputBuilder.Write("Base");
                 }
-                
+
                 outputBuilder.Write('.');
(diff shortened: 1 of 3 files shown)


ceresgalax:
@dotnet-policy-service agree

tannergooding (dismissed):

tannergooding:
Looks like there's a compile failure since Destructor is now marked nullable, which means that explicit handling needs to be added

tannergooding (approved):

Issue #461, opened 2023-07-04
Enum key stripping


dovdiienko-el3:
Could you add an option to strip the prefix from the enumeration keys. For example given following C code:

```cpp
enum abc_some_enum {
  abc_some_enum_key1,
  abc_some_enum_key2,
  abc_some_enum_key3,
};
```

The expected C# output is following:

```csharp
enum abc_some_enum {
  key1,
  key2,
  key3,
};
```

My best guess is that generator should cut maximum common prefix from the enum key name.

tannergooding:
This can be achieved via remapping today.

A new option to explicitly trim out some prefixes would also be possible, but there are many cases this won't work. This includes where the prefix is non-obvious, where it is instead a common postfix, where the first letter after the prefix is a number, etc.

I'll leave this open and up-for-grabs in case someone wants to do it, but I expect it would need to look like: `-config prefix-strip-enum`

Alternatively, or in addition, the existing `--prefixStrip` could be expanded to support specifying a common prefix to remove per enum.

fredrikhr:
In many cases the prefix is the enum member type name. So, in order to keep it simple, I suggest `--config generate-enum-member-strpped` which would simply strip the enum member type name (and following underscore) from the enum member name. That would satisfy the above example.

I am working on this now, will submit PR soon.

tannergooding:
It's worth noting one of the "best" ways to consume these types is simply to do `using static abc_some_enum;` at which point you can just use `abc_some_enum_key1` as is, unqualified.

This is ultimately very consistent with C/C++ (which can help with porting) and fits in if C# ever gets the proposed feature to allow contextual access to enum members without having to qualify the type name.

PR #527, opened 2024-02-07
Config option: strip-enum-member-type-name

fredrikhr:
Adds a new configuration option `strip-enum-member-type-name`. This feature adresses the feature request described in issue #461.

When the config option is applied, the enum type name is stripped from the beginning of each enum member.

```C
enum abc_some_enum {
  abc_some_enum_key1,
  abc_some_enum_key2,
  abc_some_enum_key3,
};
```

becomes 

```C#
enum abc_some_enum {
  key1,
  key2,
  key3,
}
```

Note above, that the original member name starts with `abc_some_enum_` which matches the enum type name `abc_some_enum` separating underscores between the type name and the suffix are also stripped to avoid leading underscores in the member names.

The matching for the enum type name is only done if the type appears verbatim in the beginning of the member name. Infix or suffix matching of the type name is not supported.

As @tannergooding notes in https://github.com/dotnet/ClangSharp/issues/461#issuecomment-1930530486, there are other solutions to make C-language stype enum be nicely usable in C#. However, the different approaches are ultimately stylistic choices and therefore I argue that this feature makes sense as a configuration option, allowing users to choose freely between styles. Following that argument, the option is kept simply to apply to the entire code generation, i.e. generation for stripping one enum, but not another is not supported as that would lead to two different styles within the same generated codebase.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
@@ -297,9 +297,7 @@ public partial class PInvokeGenerator
 
         if (Config.StripEnumMemberTypeName)
         {
-            Regex stripParentNameRegex = new($"^{Regex.Escape(parentName)}_*", RegexOptions.IgnoreCase);
-            var strippedName = stripParentNameRegex.Replace(escapedName, string.Empty);
-            escapedName = strippedName;
+            escapedName = PrefixAndStrip(escapedName, parentName, trimChar: '_');
         }
 
         var kind = isAnonymousEnum ? ValueKind.Primitive : ValueKind.Enumerator;
@@ -543,12 +541,12 @@ public partial class PInvokeGenerator
         if ((cxxMethodDecl is not null) && cxxMethodDecl.IsVirtual)
         {
             isVirtual = true;
-            escapedName = PrefixAndStripName(name, GetOverloadIndex(cxxMethodDecl));
+            escapedName = PrefixAndStripMethodName(name, GetOverloadIndex(cxxMethodDecl));
         }
         else
         {
             isVirtual = false;
-            escapedName = EscapeAndStripName(name);
+            escapedName = EscapeAndStripMethodName(name);
         }
 
         var returnType = functionDecl.ReturnType;
@@ -2032,7 +2030,7 @@ public partial class PInvokeGenerator
 
             var desc = new FunctionOrDelegateDesc {
                 AccessSpecifier = AccessSpecifier.Public,
-                EscapedName = EscapeAndStripName(name),
+                EscapedName = EscapeAndStripMethodName(name),
                 IsMemberFunction = true,
                 NativeTypeName = nativeTypeName,
                 HasFnPtrCodeGen = !_config.ExcludeFnptrCodegen,
@@ -2120,7 +2118,7 @@ public partial class PInvokeGenerator
             }
 
             var remappedName = FixupNameForMultipleHits(cxxMethodDecl);
-            var escapedName = EscapeAndStripName(remappedName);
+            var escapedName = EscapeAndStripMethodName(remappedName);
 
             var desc = new FieldDesc
             {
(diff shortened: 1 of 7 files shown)


tannergooding:
Sorry for the delay on this, I've been heads down on some dotnet/runtime related work and completely forgot to respond.

This is on my backlog (as are the other couple PRs) to finish reviewing and get them merged. At a high level glance, they all looked reasonable and I just need to sit down and finish going through all the code.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -293,6 +294,13 @@ private void VisitEnumConstantDecl(EnumConstantDecl enumConstantDecl)
             parentName = _outputBuilder.Name;
         }
 
+        if (Config.StripEnumMemberTypeName)
+        {
+            Regex stripParentNameRegex = new($"^{Regex.Escape(parentName)}_*", RegexOptions.IgnoreCase);
+            var strippedName = stripParentNameRegex.Replace(escapedName, string.Empty);
+            escapedName = strippedName;
+        }
```
tannergooding, line 302: I don't think this needs to be/should be a regex. We're really not doing anything fancy here.

This can really just be like `EscapeAndStripName` is doing, where we're just doing:
```csharp
private string EscapeAndStripEnumConstantName(string name)
{
    if (name.StartsWith(parentName, StringComparison.Ordinal))
    {
        var startIndex = parentName.Length;

        if ((name.Length > startIndex) && (name[startIndex] == '_'))
        {
            startIndex += 1;
        }

        name = name[startIndex..];
    }

    return EscapeName(name);
}
```

-- `EscapeAndStripName` should likely be renamed to `EscapeAndStripMethodName`, a similar change should be made to `PrefixAndStripName` since it also only applies to methods today.

fredrikhr: @tannergooding I have created a new general-purpose method `PrefixAndStrip` and made the enum stripping logical call that. I have also made `EscapeAndStripName` and `PrefixAndStripName` call that new method as well and renamed them as you suggested.

See commit https://github.com/dotnet/ClangSharp/pull/527/commits/6703d6a4cbbcd82557057c6443724bd929b0ba05

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -294,6 +295,11 @@ private void VisitEnumConstantDecl(EnumConstantDecl enumConstantDecl)
             parentName = _outputBuilder.Name;
         }
 
+        if (Config.StripEnumMemberTypeName)
+        {
+            escapedName = PrefixAndStrip(escapedName, parentName, trimChar: '_');
```
tannergooding, line 300: Sorry for the delay, was busy at Microsoft Build.

The rest of the changes look good, but `trimChar: '_'` is going to cause some problems and needs an extra handler. In particular we know that there exist enums of the form `DXGI_FORMAT_420_OPAQUE` and so removing the prefix `DXGI_FORMAT` + `_` will cause an invalid member name to be generated.

I'm going to merge this as is given that these are rare and there is a trivial workaround where devs can simply remap that specific member. However, if you have time it'd be great to at least add in a diagnostic when such a member is encountered and possibly add back the leading `_` to ensure we still get "valid" codegen.

tannergooding (approved):

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
fredrikhr: Sure thing, Looks like we first need a check for leading digits (i think that's the only problem from the subset of what's valid in C/C++ and C# at that point) and then another call to EscapeName in case we ended up with a keyword after stripping.
I'll make the changes.

In case of leading digits it's fine to just prepend with @ like EscapeName does?

PR #531, opened 2024-02-12
Add Declaration property to Type

ceresgalax:


tannergooding:
Sorry for the delay in looking at this.

These APIs try to mirror the general Clang APIs where possible and to not deviate from that too much. In this case, the actual `clang::Type` doesn't define any such API and instead only defines a few semi-related APIs such as `getAs*Decl` (https://clang.llvm.org/doxygen/classclang_1_1Type.html), of which ClangSharp already exposes `AsCXXRecordDecl`.

I'd be fine with also exposing the others (`AsRecordDecl`, `AsTagDecl`, `and `PointeeCXXRecordDecl`), but don't think we should expose just a `Declaration` variable to avoid any potential conflicts with future directions that Clang itself may take.

ceresgalax:
Interesting, somewhat surprising that the Clang C API has a declaration getter for types (`clang_getTypeDeclaration`) but the C++ API does not. Exposing the other getAsXXX() methods like you suggested seems to be the way to go. 

In my own project, I can write a helper method which can iterate through the different types to find a declaration, similar to how `clangsharp_Type_getDeclaration` works.

PR #535, opened 2024-04-18
ClangSharpPInvokeGenerator should exit with '0' on success

dpaoliello:
It's common practice for tools to use an exit code of 0 on success, there are various tools (like [NMake](https://learn.microsoft.com/en-us/cpp/build/reference/commands-in-a-makefile?view=msvc-170#:~:text=By%20default%2C%20NMAKE%20halts%20when%20a%20command%20returns%20a%20nonzero%20exit%20code)) that assume this.

By returning a non-zero exit code if there are warnings, and providing no way to suppress those warnings, it is difficult to integrate ClangSharpPInvokeGenerator into build systems, especially since there is no way to differentiate between "ClangSharpPInvokeGenerator emitted lots of warnings", "ClangSharpPInvokeGenerator failed to load libclang" and "dotnet failed" without inspecting stdout (as they all have positive exit codes).

### DIFF
--- a/sources/ClangSharpPInvokeGenerator/Program.cs
+++ b/sources/ClangSharpPInvokeGenerator/Program.cs
@@ -777,23 +777,9 @@ public static class Program
                     context.Console.Write("    ");
                     context.Console.WriteLine(diagnostic.ToString());
 
-                    if (diagnostic.Level == DiagnosticLevel.Warning)
+                    if (diagnostic.Level == DiagnosticLevel.Error)
                     {
-                        if (exitCode >= 0)
-                        {
-                            exitCode++;
-                        }
-                    }
-                    else if (diagnostic.Level == DiagnosticLevel.Error)
-                    {
-                        if (exitCode >= 0)
-                        {
-                            exitCode = -1;
-                        }
-                        else
-                        {
-                            exitCode--;
-                        }
+                        exitCode = -1;
                     }
                 }
             }


sources/ClangSharpPInvokeGenerator/Program.cs
```diff
@@ -777,23 +777,9 @@ public static void Run(InvocationContext context)
                     context.Console.Write("    ");
                     context.Console.WriteLine(diagnostic.ToString());
 
-                    if (diagnostic.Level == DiagnosticLevel.Warning)
+                    if (diagnostic.Level == DiagnosticLevel.Error)
```
tannergooding, line 780: This prior behavior is intentional because it did not exit with "success".

There is:
* succeeded
* exited with warnings
* exited with errors

Where error means something catastrophic happened and you should not trust the results and where warning means some kind of semi-recoverable error happened and the results are likely incorrect in more than one way.

While there is not currently a way to suppress individual warnings directly, there are options to suppress particular categories of warnings. For example you can avoid various diagnostics by using the "exclusion options": https://github.com/dotnet/ClangSharp?tab=readme-ov-file#generating-bindings. This includes both `-c ...` (for broad configuration) and `-e ...` (for explicit declaration/statement/etc exclusion).

If there are other particular classes of warnings that should be considered for exclusion, the "right" fix would be to provide new switches allowing those broad exclusions (whether it be via particular diagnostic IDs or via things that indicate that function-like macros should be ignored, for example).

Issue #536, opened 2024-04-18
How to use wildcard in `--with-using`?


dpaoliello:
The documentation for `--with-using` says:
```
A using directive to be included for the given remapped declaration name during binding generation. Supports wildcards.
```

How do wildcards work with this?
I've tried to use `--with-using=*=SomeNamespace`, but I don't see the `using` declaration for `SomeNamespace` get added.

Looking at the implementation, `PInvokeGenerator.WithUsings` doesn't set `matchStar` to true, and even if it did, `PInvokeGenerator.TryGetRemappedValue` would only match if the remapping was for `*` not arbitrary wildcards in the name.

tannergooding:
This was generally fixed by the PR. We don't strictly support "full wildcards" and that'd be a different broader feature request with a lot of expense

PR #537, opened 2024-04-18
Enable matching `*` for `--with-using`

dpaoliello:
Allows `--with-using=*=SomeNamespace` to work.

Partial fix for #536

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -7001,7 +7001,7 @@ public sealed partial class PInvokeGenerator : IDisposable
     {
         Debug.Assert(_outputBuilder is not null);
 
-        if (TryGetRemappedValue(namedDecl, _config.WithUsings, out var usings))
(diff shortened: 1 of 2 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -6945,7 +6945,7 @@ private void WithUsings(NamedDecl namedDecl)
     {
         Debug.Assert(_outputBuilder is not null);
 
-        if (TryGetRemappedValue(namedDecl, _config.WithUsings, out var usings))
+        if (TryGetRemappedValue(namedDecl, _config.WithUsings, out var usings, matchStar: true))
```
tannergooding, line 6948: Can you add a test covering this working?

dpaoliello: Done - there didn't seem to be a good place to test how options affect the output, so I created a new test class.

tannergooding (approved):

tannergooding:
Thanks!

PR #551, opened 2024-05-13
Exclude template specializations unless `generate-template-bindings` is set

dpaoliello:
ClangSharp currently excludes templates unless the `generate-template-bindings` config option is set, but it still visits template specialization declarations, resulting in the warning `Class template specializations are not supported`.

This change excludes template specializations unless the `generate-template-bindings` config option is set.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -4587,7 +4587,7 @@ public sealed partial class PInvokeGenerator : IDisposable
         bool IsExcludedByConfig(Cursor cursor)
         {
             return (_config.ExcludeFunctionsWithBody && (cursor is FunctionDecl functionDecl) && functionDecl.HasBody)
-                || (!_config.GenerateTemplateBindings && (cursor is TemplateDecl));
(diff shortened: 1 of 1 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -4587,7 +4587,7 @@ bool IsAlwaysIncluded(Cursor cursor)
         bool IsExcludedByConfig(Cursor cursor)
         {
             return (_config.ExcludeFunctionsWithBody && (cursor is FunctionDecl functionDecl) && functionDecl.HasBody)
-                || (!_config.GenerateTemplateBindings && (cursor is TemplateDecl));
+                || (!_config.GenerateTemplateBindings && ((cursor is TemplateDecl) || (cursor is ClassTemplateSpecializationDecl)));
```
tannergooding, line 4590: Some other template related declarations that don't inherit from these two include:
* `FriendTemplateDecl`
* `NonTypeTemplateParmDecl`
* `TemplateParamObjectDecl`
* `TemplateTypeParmDecl`
* `VarTemplatePartialSpecializationDecl`
* `VarTemplateSpecializationDecl`

Of these, I think `VarTemplateSpecializationDecl` and `VarTemplatePartialSpecializationDecl` are the only two that don't require first being in an already skipped templated declaration context. But it may be important to look a bit more deeply.

PR #552, opened 2024-05-21
Add support for C/C++ attribute annotate

fredrikhr:
Adds support to emit information about `__attribute__((annotate("ANNOTATION")))` in the generated Code and XML.

The GNU attribute `annotate` or its clang equivalent can appear and might convey interesting information that may be useful in the generated bindings. ClangSharpPInvokeGenerator today has only very limited support for a very few attributes, but this `annotate` attribute can be a simple catch-all allowing developers to at least have some mechanism of transferring information from the C/C++ header to the generated code.

In my particular scenario I need to do post-processing on the generated bindings and the annotation attribute seemed like a good way to transfer knowledge from the C++ header to the generated XML. I simply overrode a macro that includes the attribute in a struct type declaration, and using the changes made in this PR I could now see the emitted attribute in my C# source and as an `<attribute>` element in the generated XML. In my example I use it as a marker to change the generated struct into a ref struct, because that's what I needed in that particular case.

Similarly to `NativeTypeNameAttribute` the generator will emit a synthetically generated attribute `NativeAnnotation` if at least one `annotate` attribute was encountered. Otherwise, this new attribute is emitted under the same conditions under which `NativeTypeNameAttribute` is emitted.

In C/C++ the `annotate` attribute has one mandatory string literal argument. That literal is emitted as the `Annotation` property on the `NativeAnnotation` attribute in C#. `libClangSharp` already supports the `annotate` attribute, so this PR simply changes the P/Invoke generator to actually do something when an `annotate` attribute is reported by Clang.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -65,8 +65,6 @@ public sealed partial class PInvokeGenerator : IDisposable
     private readonly HashSet<string> _usedRemappings;
     private readonly string _placeholderMacroType;
 
-    private bool _hasAnnotateAttr;
-
     private string _filePath;
     private string[] _clangCommandLineArgs;
     private CXTranslationUnit_Flags _translationFlags;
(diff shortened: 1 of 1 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -795,6 +798,94 @@ static void GenerateNativeTypeNameAttribute(PInvokeGenerator generator, Stream?
...
         }
 
+        static void GenerateNativeAnnotationAttribute(PInvokeGenerator generator, Stream? stream, bool leaveStreamOpen)
+        {
+            var config = generator.Config;
+
+            if (!generator._hasAnnotateAttr)
+            {
+                return;
+            }
```
tannergooding, line 808: This is potentially problematic for scenarios where the developer generates helper types independently of any header that provides annotations.

This is done in various downstream tools where the tool is invoked once per header file in a library (rather than just once with all headers being processed by that single invocation).

I think it would be better to generate it by default and have a check for `--exclude NativeAnnotationAttribute` to support users who don't want it generated instead.

tannergooding:
Rest of the changes look good to me

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
fredrikhr: Good idea. I implemented this in https://github.com/dotnet/ClangSharp/pull/552/commits/c62fb6231888351dad7e8a95001f1be58421cf17

PR #563, opened 2024-07-16
Eliminate the ambiguity of "-h" option

myd7349:
Currently, the "-h" option serves as an alias for both "--headerFile" and "--help" long options, causing ambiguity.

When entering "ClangSharpPInvokeGenerator -h" in the command line, it prompts:

> Required argument missing for option: '-h'.

However, entering "ClangSharpPInvokeGenerator --help" outputs the help information correctly.

To eliminate ambiguity, we have three options:

 1. Use "-h" as the short version of "--headerFile".

    In this case, since "-h" cannot serve as the short version of "--help" anymore, we cannot rely on the default help message generated by System.Commandline. This aligns with the current behavior of "-h" as well.

 2. Use "-h" as the short version of "--help".
 
    In this case, we must change this line:

    https://github.com/dotnet/ClangSharp/blob/f0501f7370c01064676f3e8100c4fe76dc3a7c62/sources/ClangSharpPInvokeGenerator/Program.cs#L31

    either delete the "-h" alias or use a different one(`-H` or `-hf`, for example).

    (BTW, if we think `-hf` is a good option name, should we change the long version to `--header-file`?)

3. Replace `--headerFile` to another name, so that we can use a short option which is not `-h`.

This pull request adopts solution 1. Please let me know if you have any other preferences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,32 +155,33 @@ Options:
   -e, --exclude <exclude>                                          A declaration name to exclude from binding generation. []
   -f, --file <file>                                                A file to parse and generate bindings for. []
   -F, --file-directory <file-directory>                            The base path for files to parse. []
-  -h, --headerFile <headerFile>                                    A file which contains the header to prefix every generated file with. []
+  -hf, --headerFile <headerFile>                                   A file which contains the header to prefix every generated file with. []
   -i, --include <include>                                          A declaration name to include in binding generation. []
   -I, --include-directory <include-directory>                      Add directory to include search path. []
-  -x, --language <language>                                        Treat subsequent input files as having type <language>. [default: c++]
+  -x, --language <c|c++>                                           Treat subsequent input files as having type <language>. [default: c++]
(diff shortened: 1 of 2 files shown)


myd7349:
BTW, this still works:

> ClangSharpPInvokeGenerator -c h

myd7349:
TODO: Update ClangSharpPInvokeGenerator usage in https://github.com/dotnet/ClangSharp/blob/main/README.md#generating-bindings.

tannergooding:
2 is the better option here, IMO.

`-h` is a very common shorthand for help and so having `--headerFile` default to something like `-hf` is overall less problematic.

myd7349:
> 2 is the better option here, IMO.
> 
> `-h` is a very common shorthand for help and so having `--headerFile` default to something like `-hf` is overall less problematic.

Hi. Thanks. I will update this PR soon.

myd7349:
Hi. @tannergooding I have updated the PR.

The usage for ClangSharpPInvokeGenerator in README.md has also been updated to synchronize with the output of `--help` and `-c help`.

tannergooding (approved):

PR #567, opened 2024-07-30
Fix wildcard support for --with-attributes option.

ronaldvanmanen:
According to the help text the --with-attribute options supports wildcards:.
`-wa, --with-attribute <with-attribute>`

`An attribute to be added to the given remapped declaration name during binding generation. Supports wildcards. []`

However, the method `void WithAttributes(...)` invokes `TryGetMappedValue(...)` with `matchStar` set to `false` instead of `true`. This PR fixes that so that `--with-attributes *=<attribute>` will add `[<attribute>]` to all remapped declaration names during binding generation that don't have a more specific mapping (i.e. `--with-attribute X=<attribute>` will take precedence over `--with-attribute *=<attribute>`.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -6765,7 +6765,7 @@ public sealed partial class PInvokeGenerator : IDisposable
         var outputBuilder = isTestOutput ? _testOutputBuilder : _outputBuilder;
         Debug.Assert(outputBuilder is not null);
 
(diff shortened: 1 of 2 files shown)


ronaldvanmanen:
@dotnet-policy-service agree

ronaldvanmanen:
@tannergooding How do I request a review? Seems I'm not allowed to request a review for this pull request.

tannergooding (approved):

tannergooding:
Sorry for the delay, have been busy locking down the .NET 9 release and this hadn't had all CI legs passing so it didn't bubble up on my radar yet.

PR #572, opened 2024-09-23
Escape strings in `annotate` attributes

dpaoliello:
It's possible for C++ annotate attributes to contain strings literals, thus they should be correctly escaped before generating the corresponding C# attribute. The easiest way to do this is to use C# verbatim strings, and escape double-quotes (by adding a double-quote).

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
@@ -6817,7 +6817,7 @@ public sealed partial class PInvokeGenerator : IDisposable
 
                     case CX_AttrKind_Annotate:
                     {
-                        var annotationText = attr.Spelling;
+                        var annotationText = EscapeString(attr.Spelling);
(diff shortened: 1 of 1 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -6817,8 +6817,8 @@ private void WithAttributes(NamedDecl namedDecl, bool onlySupportedOSPlatform =
 
                     case CX_AttrKind_Annotate:
                     {
-                        var annotationText = attr.Spelling;
-                        outputBuilder.WriteCustomAttribute($"""NativeAnnotation("{annotationText}")""");
+                        var annotationText = attr.Spelling.Replace("\"", "\"\"", StringComparison.Ordinal);
```
tannergooding, line 6820: This should rather use the `EscapeString(string value)` helper which centralizes all the escape string handling and attempts to handle already escaped sequences as well.

dpaoliello: Done

tannergooding (approved):

PR #578, opened 2024-11-11
Fix incorrect template argument kind

MrJul:
This PR fixes wrong template argument kinds being returned from `MakeCXTemplateArgument()`.

It turns out that `TemplateArgument::ArgKind` and `CXTemplateArgumentKind` have different values, with `StructuralValue` present in the former but not the latter.

LLVM maps this explicitly: https://github.com/llvm/llvm-project/blob/00a1f1ab71302d190f8059d86a53ec62485fbce9/clang/tools/libclang/CXCursor.cpp#L1474-L1506. This PR now does the same.

I've found this issue while trying to read elements from a parameter pack (`std::tuple`). `TemplateArgument::Pack` in this case was mapped directly to `CXTemplateArgumentKind::Invalid`, making it impossible to retrieve the underlying elements. 

A unit test has been added for this case.

### DIFF
--- a/sources/libClangSharp/ClangSharp.cpp
+++ b/sources/libClangSharp/ClangSharp.cpp
@@ -26,10 +26,38 @@ using namespace clang::cxstring;
 using namespace clang::cxtu;
 using namespace clang::cxtype;
 
+CXTemplateArgumentKind ConvertTemplateArgumentKind(TemplateArgument::ArgKind kind) {
+    switch (kind) {
+    case TemplateArgument::Null:
+        return CXTemplateArgumentKind_Null;
+    case TemplateArgument::Type:
+        return CXTemplateArgumentKind_Type;
+    case TemplateArgument::Declaration:
+        return CXTemplateArgumentKind_Declaration;
+    case TemplateArgument::NullPtr:
+        return CXTemplateArgumentKind_NullPtr;
+    case TemplateArgument::Integral:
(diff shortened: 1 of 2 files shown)


MrJul:
> This is a good fix, but ideally we'd allow the devs a way to get `TemplateArgument::StructuralValue` returned as well.

Indeed, I didn't want to a new enum to avoid breaking changes.

The new test is failing on CI because `CLangSharp.Interop` directly relies on the existing libClang/libClangSharp NuGet packages, which obviously don't contain the new changes yet. While testing the changes for this PR, I was a bit surprised that there wasn't a way to target a locally built library without changing the csproj.

tannergooding:
> While testing the changes for this PR, I was a bit surprised that there wasn't a way to target a locally built library without changing the csproj.

The general issue is that building `libClangSharp` requires first having built `libClang` (there's complexity here due to the binary breaking changes libClang makes almost every release) and its quite expensive. Far more than is "reasonable" to handle on the free CI machines.

There's potentially some caching or similar that could be done, but the download would still be fairly large/expensive and its not something I've had time to investigate yet.

----------

I do need to update `libClangSharp` anyways, for the latest release of Clang, so it should be fine to just disable the test temporarily and I'll get it handled in about a week or two.

MrJul:
I've added `[Ignore]` to the test for now.

tannergooding (approved):

PR #583, opened 2024-12-10
Add support for using globs with --traverse

dpaoliello:
This adds support for using [globs](https://en.wikipedia.org/wiki/Glob_(programming)) with the `--traverse` option, via the [`DotNet.Glob` library](https://github.com/dazinator/DotNet.Glob).

I initially attempted to use the `Microsoft.Extension.FileSystemGlobbing` library, however it doesn't support globs that are absolute paths (see <https://github.com/dotnet/runtime/issues/62333>), hence that would break any existing use of `--traverse` that was using an absolute path.


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.cs
```diff
@@ -4908,26 +4924,34 @@ bool IsExcludedByName(Cursor cursor, ref uint isExcludedValue)
...
+            // If there is no source file, then exclude it.
+            if (string.IsNullOrEmpty(fileName))
             {
-                return true;
+                return false;
             }
-            else if (_config.TraversalNames.Contains(fileName.NormalizeFullPath(), equalityComparer))
+
+            // If the user has specified a set of headers to include, then use that.
+            if (_traversalGlobs.Count > 0)
```
tannergooding, line 4941: I'm not really sold on the globbing support here, particularly given how other parameters and matching is being done. There's kind of 3 parts to it here...

1. ClangSharp itself is being signed and used by tooling where the full trust/validation chain is important. Dotnet.Glob hasn't been touched in 4ish years now, while there are several newer issues opened; so its not clear if it's a good long term thing to depend on for a project like this

2. It's somewhat intentional that traversal be more explicit to help with determinism, various file system issues (like reparse points or symbolic links) and to avoid accidentally bringing in "everything" which causes conflicts across multiple invocations of the tool. There's also some complexities in how Clang itself surfaces file paths that make it a bit error prone and doesn't always work with file system APIs.

3. All the other command line parameters are using `*` very simplistically or with `Regex`, so this creates a mismatch with the other handling. I imagine the basic support that's needed here could itself be added relatively simply or via an alternative command line option; such as if you want to include all files under a given folder then a `--traverseDirectory` might be a better option

dpaoliello:
Abandoning: looks like I can do what I need to without having to add glob support directly to ClangSharp.

PR #585, opened 2024-12-13
Ignore the NoInline attribute

dpaoliello:
The C++ NoInline attribute doesn't map to anything meaningful for C# or P/Invoke, so it can safely be ignored.


tannergooding (approved):

tannergooding:
> The C++ NoInline attribute doesn't map to anything meaningful for C# or P/Invoke, so it can safely be ignored.

A small nit is that it might be relevant for non P/Invoke methods to bind as `[MethodImpl(MethodImplOptions.NoInlining)]`, but that's relatively uncommon.

PR #590, opened 2025-01-08
Add Parameter Types to InheritDoc comments

curin:
InheritDoc comments were ambiguous between different overrides so this will specify which override we are to inherit from.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -372,6 +372,9 @@ internal partial class CSharpOutputBuilder : IOutputBuilder
                 Write(desc.ParentName);
                 Write('.');
                 Write(desc.EscapedName);
+                Write('(');
+                Write(string.Join(", ", desc.ParameterTypes));
+                Write(')');
                 WriteLine("\" />");
             }
             else
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -372,7 +372,7 @@ internal partial class CSharpOutputBuilder : IOutputBuilder
                 Write(desc.ParentName);
                 Write('.');
                 Write(desc.EscapedName);
-                if (desc.ParameterTypes is not null && desc.ParameterTypes.Length > 0)
+                if (desc.ParameterTypes is not null)
                 {
                     Write('(');
                     Write(string.Join(", ", desc.ParameterTypes));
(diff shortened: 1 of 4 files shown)


sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
```diff
@@ -372,6 +372,9 @@ public void BeginFunctionOrDelegate(in FunctionOrDelegateDesc desc, ref bool isM
                 Write(desc.ParentName);
                 Write('.');
                 Write(desc.EscapedName);
+                Write('(');
+                Write(string.Join(", ", desc.ParameterTypes));
+                Write(')');
```
tannergooding, line 377: nit: It would be beneficial to only do this when there are potential conflicts, it's a lot of extra text that's unnecessary in most cases (particularly for regular C exports, since its really only applicable to C++ exports).

There's some existing helper functions that determine if multiple members with the same name exist that can likely be used.

curin: fair, do you know where those functions live off hand?

tannergooding: @curin sorry, thought I had already responded to this.

You likely want to reuse a similar approach to `FixupNameForMultipleHits`: https://source.clangsharp.dev/#ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs,dbbe5069d6d4e556,references

It might involve tweaking `GetOverloadIndex` slightly to allow disambiguating "no overloads" (-1) from "some overloads" (>= 0), where today we just track a raw index instead not how many overloads existed.

curin: I ended up adding a function just to get the overload count based on `GetOverloadIndex`. The only difference is that I doesn't cache the count like indicies do. I can add that if you want, but wasn't going to add the memory overhead unless you thought it was a good idea.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -501,12 +501,15 @@ private void VisitFunctionDecl(FunctionDecl functionDecl)
...
         var cxxMethodDecl = functionDecl as CXXMethodDecl;
+        uint overloadCount = 0;
 
         if (cxxMethodDecl is not null and CXXConstructorDecl)
         {
             var parent = cxxMethodDecl.Parent;
             Debug.Assert(parent is not null);
             name = GetRemappedCursorName(parent);
+
+            overloadCount = GetOverloadCount(cxxMethodDecl);
```
tannergooding, line 512: Is this sufficient?

This will only set overload count for constructor decls, when presumably we want it for all cxxMethodDecl instead.

curin: you are absolutely right and I read that entirely wrong. fixed now.

tannergooding (approved):

PR #594, opened 2025-03-14
Add DontUseUsingStaticsForGuidMember

curin:
This causes some issues with some renaming in Silk.NET and would like to toggle it off.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
@@ -267,6 +267,8 @@ public sealed class PInvokeGeneratorConfiguration
 
     public bool StripEnumMemberTypeName => _options.HasFlag(PInvokeGeneratorConfigurationOptions.StripEnumMemberTypeName);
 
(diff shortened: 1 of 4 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
```diff
@@ -267,6 +267,8 @@ public IReadOnlyCollection<string> ExcludedNames
 
     public bool StripEnumMemberTypeName => _options.HasFlag(PInvokeGeneratorConfigurationOptions.StripEnumMemberTypeName);
 
+    public bool DontUseUsingStaticsForGuidMember => _options.HasFlag(PInvokeGeneratorConfigurationOptions.DontUseUsingStaticsForGuidMember);
```
Perksey, line 270: There is some corresponding code in Program.cs of ClangSharpPInvokeGenerator that needs updating to actually set this flag (Silk.NET's ResponseFileHandler is a manually maintained subset of this Program.cs file)

Perksey (changes requested):

sources/ClangSharp.PInvokeGenerator/PInvokeGeneratorConfiguration.cs
curin: good catch, fixed it

Perksey (approved):

tannergooding (approved):

PR #595, opened 2025-04-27
fix InvalidCastException when inheriting c++ struct constructors

lishinko:
while generating binding file from [Viewport.h](https://github.com/google/filament/blob/main/filament/include/filament/Viewport.h)  from Google Filament Project, an InvalidCastException occured and the generation process exited. 
The reason is : C# dosn't support constuctor inheriting, so we cannot generate correct code for it. 
This PR just bypasses the exception and let the generation continue. The generated file is wrong, but it is caused by the struct inerhitance, not by this exception any more.


lishinko:
@dotnet-policy-service agree

Issue #611, opened 2025-07-09
ClangSharpPInvokeGenerator does not handle unsigned int bitshifts correctly


Exanite:
When generating bindings for Vulkan using ClangSharpPInvokeGenerator through Silk.NET 3, the following C code is not converted into compilable C# code due to the bitshifts.

This results in the following error:
```
/Vulkan/Vulkan.gen.cs(25934,10): error CS0019: Operator '<<' cannot be applied to operands of type 'uint' and 'uint'
```

---

Relevant C code: https://github.com/KhronosGroup/Vulkan-Headers/blob/main/include/vulkan/vulkan_core.h
```c
#define VK_MAKE_API_VERSION(variant, major, minor, patch) \
    ((((uint32_t)(variant)) << 29U) | (((uint32_t)(major)) << 22U) | (((uint32_t)(minor)) << 12U) | ((uint32_t)(patch)))


//#define VK_API_VERSION VK_MAKE_API_VERSION(0, 1, 0, 0) // Patch version should always be set to 0

// Vulkan 1.0 version number
#define VK_API_VERSION_1_0 VK_MAKE_API_VERSION(0, 1, 0, 0)// Patch version should always be set to 0
```

Code as generated by Silk.NET:
```cs
[NativeTypeName("#define VK_API_VERSION_1_0 VK_MAKE_API_VERSION(0, 1, 0, 0)")]
public const uint VK_API_VERSION_1_0 = (
    (((uint)(0)) << 29U) | (((uint)(1)) << 22U) | (((uint)(0)) << 12U) | ((uint)(0))
);
```

Code as found in TerraFX Vulkan: https://raw.githubusercontent.com/terrafx/terrafx.interop.vulkan/refs/heads/main/sources/Interop/Vulkan/Vulkan/vulkan/vulkan_core/Vulkan.cs
```cs
[NativeTypeName("#define VK_API_VERSION_1_0 VK_MAKE_API_VERSION(0, 1, 0, 0)")]
public const uint VK_API_VERSION_1_0 = ((((uint)(0)) << 29) | (((uint)(1)) << 22) | (((uint)(0)) << 12) | ((uint)(0)));
```

Code when I regenerated the TerraFX Vulkan bindings locally (my Windows SDK version and Vulkan SDK version don't match, but I don't think that affects the code in question):
Note that the output matches Silk.NET, except in formatting, and the right-hand side of the bitshifts are unsigned ints.
```cs
[...]

PR #612, opened 2025-07-10
Fix compile error when generating code involving unsigned int bitshifts

Exanite:
This fixes https://github.com/dotnet/ClangSharp/issues/611.

~~This changes it so that the right-hand side of bitshifts always gets casted to an int.
This works because C# only accepts ints as the RHS of bitshifts: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1211-shift-operators~~

~~Note: This leads to widespread changes in how the bitshift operator is output (see test cases). If necessary, I can look into a less invasive approach. The reasoning for doing this globally is to keep things consistent and the extra cast shouldn't affect anything other than adding a bit of bloat to the generated code.~~

This now casts only if the RHS is a non-literal, non-`int` expression. Otherwise, the literal or expression is used directly.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
@@ -42,7 +42,7 @@ public partial class PInvokeGenerator
         outputBuilder.Write(binaryOperator.OpcodeStr);
         outputBuilder.Write(' ');
 
-        if (binaryOperator.IsShiftOp || binaryOperator.IsShiftAssignOp)
+        if ((binaryOperator.IsShiftOp || binaryOperator.IsShiftAssignOp) && binaryOperator.RHS.Type.Kind != CXType_Int)
         {
             // RHS of shift operation in C# must be an int
             outputBuilder.Write("(int)");
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
@@ -41,7 +41,22 @@ public partial class PInvokeGenerator
         outputBuilder.Write(' ');
         outputBuilder.Write(binaryOperator.OpcodeStr);
         outputBuilder.Write(' ');
-        Visit(binaryOperator.RHS);
+
+        if (binaryOperator.IsShiftOp || binaryOperator.IsShiftAssignOp)
+        {
+            // RHS of shift operation in C# must be an int
+            outputBuilder.Write("(int)");
+            outputBuilder.Write('(');
+
+            Visit(binaryOperator.RHS);
+
+            outputBuilder.Write(')');
+        }
+        else
+        {
+            Visit(binaryOperator.RHS);
+        }
+
         StopCSharpCode();
     }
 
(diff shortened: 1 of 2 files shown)


tests/ClangSharp.PInvokeGenerator.UnitTests/Base/FunctionDeclarationBodyImportTest.cs
```diff
@@ -28,11 +28,7 @@ public abstract class FunctionDeclarationBodyImportTest : PInvokeGeneratorTest
     [TestCase("-=")]
     [TestCase("/")]
     [TestCase("/=")]
-    [TestCase("<<")]
-    [TestCase("<<=")]
     [TestCase("=")]
-    [TestCase(">>")]
-    [TestCase(">>=")]
```
Exanite, line 35: I'm not sure if removing the bitshift operators from the BinaryOperatorBasicTest case is a good idea, but I don't know how else to specify the difference in behavior. Let me know what you think.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -41,7 +41,22 @@ private void VisitBinaryOperator(BinaryOperator binaryOperator)
...
+
+        if (binaryOperator.IsShiftOp || binaryOperator.IsShiftAssignOp)
+        {
+            // RHS of shift operation in C# must be an int
+            outputBuilder.Write("(int)");
+            outputBuilder.Write('(');
+
+            Visit(binaryOperator.RHS);
+
+            outputBuilder.Write(')');
```
tannergooding, line 53: I wouldn't do this so that it always casts. Rather, I would attempt to handle some common cases like `constant integers` and values that are already of the right type to not need the cast

We should really only cast if it is an unknown variable of type `uint`, `long`, `ulong`, etc.

Exanite: Thanks for the review and pointers! I'll take a look

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -41,7 +41,32 @@ private void VisitBinaryOperator(BinaryOperator binaryOperator)
...
+        {
+            // RHS of shift operation in C# must be an int
+            if (binaryOperator.RHS.Type.Kind is CXType_Int or CXType_Long)
+            {
+                Visit(binaryOperator.RHS);
+            }
+            else if (binaryOperator.RHS is IntegerLiteral literal)
+            {
+                outputBuilder.Write(literal.Value);
+            }
```
tannergooding, line 55: This won't do the right thing for an integer literal whose value is actually larger than 32-bits. Which while unlikely, should probably be handled.

tannergooding:
Mostly looks right, just a couple of edge cases that I think were missed

Exanite:
I think I got most of the cases.
The tested output all compiles:
image

I still need to check how the generator behaves when `GenerateDisableRuntimeMarshalling` and/or `GenerateUnixTypes` is set to true.

There are also a few unideal cases that require more work than it is probably worth.
I note these in the test cases:
```cs
// Non-ideal cases:
// UByte
// IntMin
// ULongMax
// Hexadecimal
```

Note that the unchecked is getting added by some other part of the generator. I did not modify that behavior.

tannergooding (approved):

PR #613, opened 2025-07-14
Add regex support to remapping and attributes

TareHimself:
This PR does not include tests. I created it to determine if the repo will allow regex for `--remap` and `--with-attributes`. regex support is opt-in using a separate `--remap-regex` and `--with-attributes-regex` , `--remap-regex` supports string formatting. These are some examples
```
--remap-regex
^([\w\d]+)Flags([0-9])$={0}Flags{1}
^([\w\d]+)FlagBits$=@{0}Flags

--with-attribute-regex
^Vk[\w\d]+Flags$=Flags
^Vk[\w\d]+Flags[0-9]$=Flags
```
if allowed this can also be added to `--with-type` as `--with-type-regex`


TareHimself:
@dotnet-policy-service agree

PR #626, opened 2025-08-21
Fix missing type casts when using bitfields with remapped types (and some other cases)

### LINKED ISSUE #618: Missing type casts when using bitfields with remapped types
This is an edge case that occurs when using remappings with bitfields.

I discovered this issue when attempting to map Vulkan `Flags` enums to `FlagBits` in order to have ClangSharpPInvokeGenerator output `FlagBits` as the property type rather than `uint`.

One example of this issue happening is when generating the struct, `VkAccelerationStructureInstanceKHR`, with the mapping, `VkGeometryInstanceFlagsKHR` -> `VkGeometryInstanceFlagBitsKHR`. (Search for the struct name in [vulkan_core.h](https://raw.githubusercontent.com/KhronosGroup/Vulkan-Headers/refs/heads/main/include/vulkan/vulkan_core.h))

This is a simplified example that I implemented as a test case:
```cs
typedef enum FlagBits {
    Member = 0x7FFFFFFF
} FlagBits;
typedef unsigned int Flags;

typedef struct Bitfield {
    uint bits : 8;
    Flags flags : 8;
} Bitfield;
```

This outputs the following, without the proper casts. This causes a compile error.
The expected is that the get return value is cast to `FlagBits` and `value` is cast to `uint`.
```cs
public uint _bitfield;

[NativeTypeName("Flags : 8")]
public FlagBits flags
{
    readonly get
    {
        return (_bitfield >> 8) & 0xFFu;
    }

    set
    {
        _bitfield = (_bitfield & ~(0xFFu << 8)) | ((value & 0xFFu) << 8);
    }
}
```

---

You can find the full test cases here in my fork: <https://github.com/Exanite/ClangSharp/blob/d3a1dc1a90b85ba465f15b72f60992cbe4c28be7/tests/ClangSharp.PInvokeGenerator.UnitTests/CTest.cs#L456>

Note that `BitfieldEnumPropertyTypeCastTest` **already passes**.
`BitfieldEnumPropertyTypeCastWithRemappingTest` is the test that fails.

This shows that ClangSharp already handles this case to some extent and that it is the remapping that causes issues.

---

If this is a case that ClangSharp should handle, I'm interested in opening a PR for fixing this issue.

Exanite:
This PR fixes https://github.com/dotnet/ClangSharp/issues/618 and some other issues that I identified as I was looking to understand the code.

https://github.com/dotnet/ClangSharp/issues/618 explains the two main cases fixed (missing type casts when remapping and incorrect sbyte case).

I also changed the `uint/int` case, mainly as an implementation detail. It doesn't matter if the cast happens before or after the left shift. It just has to happen before the right shift. This keeps the behavior consistent across cases.
Edit: I only changed the comment. The previous behavior already matches the current and the comment.

```diff
  // backing  int, current int            (value << cns) >> cns
  // backing  int, current uint           (uint)((value >> cns) & msk)

- // backing uint, current int            ((int)value << cns) >> cns
+ // backing uint, current int            (int)(value << cns) >> cns
  // backing uint, current uint           (value >> cns) & msk

  // backing uint, current byte           (byte)((value >> cns) & msk)
- // backing uint, current sbyte          (sbyte)((value << cns) >> cns)
+ // backing uint, current sbyte          (sbyte)((sbyte)(value << cns) >> cns)
```

You'll notice the code has changed a significant amount. This is because I rewrote most of the conditions and grouped the code together to make it a bit more readable. Because it was hard to track scopes, such as where each matching opening/closing parentheses were located, I added variables identifying the purpose and conditions for each scope.

Remaining tasks
- [x] Separate tests into Unix/Windows variants. I've tested on both platforms by manually accounting for the platform differences (different bitfield packing; different enum backing types), but I haven't added tests for both platforms yet.
- [x] Continue verifying that my new implementation works. It currently passes all test cases, but because of the larger changes made and how many conditions are required, I'm not yet confident that every case works.
   - Looks like CI tests failed. Windows tests failing is expected. Linux ARM tests failing is unexpected, but shouldn't be too hard to fix. I just need to research how C/C++ on Linux ARM behaves. Specifically, if `char` is a `byte` or `sbyte` on Linux ARM.
- [x] Update branch with changes from main


Exanite:
Ready for review.
All test cases pass.
Comparing current behavior with previous behavior shows that only the two targeted cases have been changed.

Since I changed the indentation, it's easier to see the changes if you ignore whitespace in your diff viewer.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -2748,204 +2748,282 @@ void VisitBitfieldDecl(FieldDecl fieldDecl, BitfieldDesc[] bitfieldDescs, Record
...
-            var isUnsignedToSigned = !isTypeBackingSigned && isTypeSigned;
+            // Small types become uint32/int32s after shifting
+            var isSmallType = fieldDecl.Type.Handle.SizeOf < 4;
+            var isSmallTypeBacking = currentSize < 4;
 
-            var needsCast = isSmallType || isRemappedToSelf || isTypeMismatch || isUnsignedToSigned;
-            var needsParenFirst = !isSmallType && isUnsignedToSigned;
-            var needsParenSecond = !needsParenFirst || isRemappedToSelf;
+            // Check if field/backing types match
+            var isTypeMismatch = type != typeBacking;
```
tannergooding, line 2759: Why the comparison against `typeBacking` rather than `buildinTypeBacking`? Won't that be meaningful for certain typedefs or enums?

tannergooding (dismissed):
Overall LGTM. Just a minor nit and a question that I don't remember the answer to off the top of my head

Exanite:
Good catch. Using `buildinTypeBacking` instead of `typeBacking` is in fact important.
I've added a test case and changed the code to use `buildinTypeBacking` instead.

The difference with `typeBacking` is that an unnecessary type cast gets inserted.

tannergooding (approved):

PR #634, opened 2025-09-27
proposal: Add documents for library use

aka-nse:
related: #561

----

Hello.

I tried to use ClangSharp as a library, but found the documentation lacking, which made learning quite challenging.
Looking at the discussions, it seems others have faced the same issue.

Therefore, I would like to propose adding an introduction for library usage.
I've written some content myself and will submit it as a PR.
While a Wiki would normally be more suitable for this purpose, I've added a `docs/` directory to make it submittable as a PR.

I would appreciate your feedback and suggestions.


tannergooding:
Closing, replaced by https://github.com/dotnet/ClangSharp/pull/754

Issue #607, opened 2025-06-17
Support C++ single-quote digit separators


jwosty:
For example, see this C++ code for which I am trying to generate bindings: https://github.com/facebook/yoga/blob/be00354b7194bf182c9b66584f64c83e248d3007/yoga/style/StyleValueHandle.h#L55

Would be nice if the P/Invoke generator supported this syntax correctly. Currently it emits the literal as-is (with single quotes in it), but it should either strip those out, or if you want to get fancy, replace them with C#'s own digit separator (`_`).

PR #639, opened 2025-10-12
feat: support single quote digits separator

davidelettieri:
This PR adds support for single quote digits separator by converting into the underscore supported by C#.

Close https://github.com/dotnet/ClangSharp/issues/607

A couple of things I noticed while working on this:
- some files are missing the copyright line at the top. I added it to the new file I created but I avoided changes to other files.
- the following code is translated into a class with a private field but without initialization. If there is interest in having the initialization in the generated C# code, I'll be happy to work on the implementation.
```cpp
class MyClass
{
public:
	float x = 1024;
};
```

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.cs
@@ -132,18 +132,16 @@ internal sealed partial class CSharpOutputBuilder(string name, PInvokeGenerator 
 
     public void WriteNumberLiteral(ReadOnlySpan<char> value)
     {
-        for (var i = 0; i < value.Length; i++)
+        var index = value.IndexOf('\'');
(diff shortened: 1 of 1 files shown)


sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitStmt.cs
```diff
@@ -1201,12 +1201,12 @@ private void VisitFloatingLiteral(FloatingLiteral floatingLiteral)
         var outputBuilder = StartCSharpCode();
         if (floatingLiteral.ValueString.EndsWith(".f", StringComparison.Ordinal))
         {
-            outputBuilder.Write(floatingLiteral.ValueString.AsSpan()[..^1]);
```
tannergooding: The `.AsSpan()` here is important, it avoids unnecessarily allocating for the common cases, which helps keep GC pressure down and performance up

PR #646, opened 2025-10-28
Update testing framework

ds5678:
Merge duplicate code under a new class `CXXMethodDeclarationCSharpTest`


tannergooding:
> Every test method has a pair of files (corresponding to input and output) and a set of multiple options on which to test.

I am not a fan of checking in what are effectively "baseline" files or building up complex infrastructure (which effectively needs its own testing) to run these tests.

The current approach had been done with because it was simple. What was largely missing was the sharing of the test between versions when it was identical (what this PR largely fixes) to avoid unnecessary duplication. 

There is then another factor where the `inputContents` are always identical between tests and its only the `expectedOutputContents` that should differ in some cases. Sharing the input is fairly easy while tweaking the output between the tests is harder without more error prone infrastructure.

Part of it could potentially be done by just having clear hierarchy to the tests. `CXXMethodDeclarationTest` and then a `Compatible` which inherits from that and a `Preview : Latest : Compatible` setup. Windows vs Unix differences would probably have to be changed to be based off a condition since you can't have multiple inheritance. This would maximize sharing and only require overriding in the case where the newer deviates from the older.

tannergooding (approved):

PR #648, opened 2025-10-30
Make marker interfaces partial to support user extensions

### LINKED ISSUE #647: Marker interfaces should be partial to support user extensions
## Problem

When `--generate-marker-interfaces` is enabled, generated nested interfaces are not marked as `partial`, preventing users from manually extending them with custom implementations.

### Simple Example

**C++ input:**
```cpp
struct IFoo
{
    virtual void Method1() = 0;
    virtual void Method2() = 0;
};
```

**Generated with `--generate-marker-interfaces`:**
```csharp
public unsafe partial struct IFoo : IFoo.Interface
{
    public interface Interface  // NOT partial - users cannot extend
    {
        void Method1();
        void Method2();
    }
}
```

**User tries to manually add a custom helper method:**
```csharp
// User's manually-written code in a separate file
public partial struct IFoo
{
    public partial interface Interface  // ✗ Error: no partial declaration exists
    {
        void CallBoth();  // User wants to add this to the interface contract
    }
    
    // User's manual implementation
    public void CallBoth() 
    {
        Method1();
        Method2();
    }
}
```

**Generic constraint breaks:**
```csharp
void Process<T>(T foo) where T : unmanaged, IFoo.Interface
{
    foo.CallBoth();  // ✗ Error: Interface doesn't contain CallBoth
}
```

## Solution

Change `CSharpOutputBuilder.VisitDecl.cs:829`:
```csharp
WriteIndented("public partial interface Interface");
```

This allows users to manually extend the interface with custom methods while maintaining backward compatibility (adding `partial` is non-breaking).

I have a fix ready in my fork.

tompipe:
## Summary

Makes generated marker interfaces `partial` to allow users to manually extend them with custom implementations.

Fixes #647

## Changes

- Modified `CSharpOutputBuilder.VisitDecl.cs:829` to generate `public partial interface Interface` instead of `public interface Interface`

## Why

When `--generate-marker-interfaces` is enabled, users may want to manually add custom helper methods to the interface contract. Without the `partial` keyword, this is impossible.

## Impact

- **Non-breaking**: Adding `partial` to an interface is backward compatible
- **Consistent**: Matches the existing pattern where generated structs are already `partial`
- **Enables**: User extensions for excluded methods, custom helpers, or manual implementations of unsupported C++ constructs

## Testing

- Built successfully with `dotnet build`
- No compilation errors or warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
+++ b/sources/ClangSharp.PInvokeGenerator/CSharp/CSharpOutputBuilder.VisitDecl.cs
@@ -826,7 +826,7 @@ internal partial class CSharpOutputBuilder : IOutputBuilder
 
     public void BeginMarkerInterface(string[]? baseTypeNames)
     {
-        WriteIndented("public interface Interface");
+        WriteIndented("public partial interface Interface");
 
         if (baseTypeNames is not null)
         {
--- a/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/CXXMethodDeclarationTest.cs
+++ b/tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/CXXMethodDeclarationTest.cs
@@ -633,7 +633,7 @@ namespace ClangSharp.Test
             }}
         }}
 
-        public interface Interface
+        public partial interface Interface
         {{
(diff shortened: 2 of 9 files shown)


tompipe:
@dotnet-policy-service agree

PR #675, opened 2026-01-06
Add FloatingLiteral.ValueAsFloat and ValueAsDouble.

rolfbjarne:


sources/libClangSharp/ClangSharp.cpp
```diff
@@ -1367,6 +1367,38 @@ double clangsharp_Cursor_getFloatingLiteralValueAsApproximateDouble(CXCursor C)
     return 0;
 }
 
+unsigned clangsharp_Cursor_getFloatingLiteralValueAsDouble(CXCursor C, double* value) {
```
tannergooding: Not sure I understand the need for these APIs?

`ValueAsApproximateDouble` only loses data in the case that we have some "more precise" type than `double`, such as `x87DoubleExtended` or `IEEEquad` (which aren't portable to C# right now anyways).

So what it looks like you want is just some `unsigned clangsharp_Cursor_isFloatingLiteralRepresentableBy(CXCursor C, CX_FloatingSemantics semantics)`.

Such an API then allows this to be more generally supported in managed land instead, as you can simply check "is the floating-literal regardless of semantics losslessly representable by this other semantic". You can then take and cast `ValueAsApproximateDouble` to the target format in that case for the two types that are being supported here.

PR #679, opened 2026-01-07
Improve performance when fetching Decls.

rolfbjarne:
The current implementation in clangsharp_Cursor_getDecl to get a Decl is an N^2 algorithm; getting a Decl for index X means looping through all the previous indices 0-(X-1) first.

This is rather slow when dealing with hundreds of thousands of Decl instances, here's a screenshot from Instruments showing 74% of the time spent in clangsharp_Cursor_getDecl:

Screenshot 2026-01-07 at 12 47 26

So I added a way in LazyList to use an existing item to get the next item, and then use Decl.NextDeclInContext to take advantage of this, effectively making an N algorithm (which becomes 2N, because we iterate over all the Decls first to count them).

My main scenario now runs in ~51 seconds instead of 3m05 seconds, so less that 1/3 of the time.

### DIFF
--- a/sources/ClangSharp/LazyList`1.cs
+++ b/sources/ClangSharp/LazyList`1.cs
@@ -42,7 +42,9 @@ internal sealed class LazyList<[DynamicallyAccessedMembers(DynamicallyAccessedMe
                 if (_valueFactoryWithPreviousValue is not null)
                 {
                     item = _valueFactoryWithPreviousValue(index, index == 0 ? null : _items[index - 1]);
-                } else {
+                }
+                else
+                {
                     item = _valueFactory!.Invoke(index);
                 }
                 items[index] = item;


sources/ClangSharp/Cursors/Decls/Decl.cs
```diff
@@ -35,7 +35,29 @@ private protected Decl(CXCursor handle, CXCursorKind expectedCursorKind, CX_Decl
         _attrs = LazyList.Create<Attr>(Handle.NumAttrs, (i) => TranslationUnit.GetOrCreate<Attr>(Handle.GetAttr(unchecked((uint)i))));
         _body = new ValueLazy<Stmt?>(() => !Handle.Body.IsNull ? TranslationUnit.GetOrCreate<Stmt>(Handle.Body) : null);
         _canonicalDecl = new ValueLazy<Decl>(() => TranslationUnit.GetOrCreate<Decl>(Handle.CanonicalCursor));
-        _decls = LazyList.Create<Decl>(Handle.NumDecls, (i) => TranslationUnit.GetOrCreate<Decl>(Handle.GetDecl(unchecked((uint)i))));
```
rolfbjarne: FWIW counting the decls (i.e. calling `Handle.NumDecls`) also shows up in the profile, although not nearly as prominent, so I also considered lazy-loading the count of LazyList instead of having to specify it up-front.

tests/ClangSharp.UnitTests/CXCursor.cs
```diff
@@ -10,6 +10,7 @@ public class CXCursorTest : TranslationUnitTest
     [Test]
     public void AttrKindSpelling()
     {
+        AssertNeedNewClangSharp();
```
rolfbjarne: The new native libClangSharp method makes a lot of tests fail because the new code is in a rather fundamental part of ClangSharp, not sure how to handle this?

tannergooding: The simplest thing would be to use `NativeLibrary.GetExport` to try and resolve the new `GetDecls` API and fallback to the old logic if not present.

I can then fix that up to directly use the new API when I get the new libClangSharp published.

sources/ClangSharp/LazyList`1.cs
```diff
@@ -30,7 +39,12 @@ public T this[int index]
 
             if (item is null)
             {
-                item = _valueFactory(index);
+                if (_valueFactoryWithPreviousValue is not null)
+                {
+                    item = _valueFactoryWithPreviousValue(index, index == 0 ? null : _items[index - 1]);
+                } else {
```
tannergooding:
```suggestion
                }
                else
                {
```

tannergooding (approved):

PR #699, opened 2026-06-29
Fix generation of long-typed defines when on Linux

Exanite:
## Summary

For context, C `long` is output as C# `nint` on Linux.

This causes an issue where a 64-bit `ulong` value can be used as the initializer for a `const` field of type `nint`, leading to 2 errors:
1. There is a missing type cast from `long` to `nint`, causing a compile error.
2. Because C# doesn't know the bit width of `nint` until run time, this errors with `Constant initializer must be compile-time constant` after fixing the first error. This is specified here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/native-integers.md

This PR fixes the issue by updating the generator to output the expected test output below, along with some additional test cases.

## Current test outputs (matching main branch)

This is the current output of the new `CLongDefinesTestUnix` test case on Linux when ran against the main branch.

```cs
public static partial class Methods
{
    [NativeTypeName("#define SIZE_MAX (18446744073709551615UL)")]
    public const nuint SIZE_MAX = (18446744073709551615U);

    [NativeTypeName("#define CL_IMPORT_MEMORY_WHOLE_ALLOCATION_ARM SIZE_MAX")]
    public const nuint CL_IMPORT_MEMORY_WHOLE_ALLOCATION_ARM = (18446744073709551615U);

    [NativeTypeName("#define LONG_MAX __LONG_MAX__")]
    public const nint LONG_MAX = unchecked(9223372036854775807);

    [NativeTypeName("#define ULONG_MAX (__LONG_MAX__ *2UL+1UL)")]
    public const nuint ULONG_MAX = (9223372036854775807 * 2U + 1U);
}
```

The new `CLongDefinesTestWindows` test case already passes and is there to ensure that no unintended changes occur to the Windows behavior.

## Expected test outputs

The expected test outputs are my proposal for *roughly* what the expected bindings output should look like.

Notably, we need to add a type cast and change the field to be a `static readonly` field instead of a `const` field.

```cs
public static partial class Methods
{
    [NativeTypeName(""#define SIZE_MAX (18446744073709551615UL)"")]
    public static readonly nuint SIZE_MAX = unchecked((nuint)(18446744073709551615U));

    [NativeTypeName(""#define CL_IMPORT_MEMORY_WHOLE_ALLOCATION_ARM SIZE_MAX"")]
    public static readonly nuint CL_IMPORT_MEMORY_WHOLE_ALLOCATION_ARM = unchecked((nuint)(18446744073709551615U));

    [NativeTypeName(""#define LONG_MAX __LONG_MAX__"")]
    public static readonly nint LONG_MAX = unchecked((nint)(9223372036854775807));

    [NativeTypeName(""#define ULONG_MAX (__LONG_MAX__ *2UL+1UL)"")]
    public static readonly nuint ULONG_MAX = unchecked((nuint)(9223372036854775807 * 2U + 1U));
}
```

## Remaining Tasks

I'll handle these tasks before undrafting.

- [x] Add tests for the "similar" case in my comment below
- [x] Identify other test cases to add
- [x] Address the todo comments I left
- [x] Package this branch and test against Silk.NET 3.0 on both Windows and Linux
   - [x] Linux
   - [x] Windows
- [x] Add test case for `SDL_size_add_check_overflow` ~~and fix unnecessary `unchecked` statement~~
   - Decided to defer the fix. The issue is tracked here: https://github.com/dotnet/ClangSharp/issues/709


Exanite:
There is also a similar case with `nint` typed `const` fields:

Also from OpenCL:
```c
#if INTPTR_MAX == INT32_MAX
#define CL_ICD2_TAG_KHR ((intptr_t)0x434C3331)
#else
#define CL_ICD2_TAG_KHR ((intptr_t)0x4F50454E434C3331)
#endif
```

This generates as the following on both Windows and Linux:
```cs
[NativeTypeName("#define CL_ICD2_TAG_KHR ((intptr_t)0x4F50454E434C3331)")]
public const nint CL_ICD2_TAG_KHR = unchecked((nint)(0x4F50454E434C3331));
```

In this case, the only change needed to make this compile would be to replace `const` with `static readonly`.

tests/ClangSharp.PInvokeGenerator.UnitTests/CSharpCompatibleUnix/VarDeclarationTest.cs
```diff
@@ -267,7 +267,7 @@ namespace ClangSharp.Test
     public static partial class Methods
     {{
         [NativeTypeName(""#define MyMacro1 (long)0x80000000L"")]
-        public const IntPtr MyMacro1 = unchecked((nint)(0x80000000));
```
Exanite, line 270: Old expected output doesn't actually compile and fails with `The expression being assigned to 'MyMacro1' must be constant`.

This is easy to miss since the error does not show until you actually compile.

sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.VisitDecl.cs
```diff
@@ -2587,7 +2587,7 @@ void VisitBitfieldDecl(FieldDecl fieldDecl, BitfieldDesc[] bitfieldDescs, Record
 
                 case CXType_ULongLong:
                 {
-                    if (typeNameBacking.Equals("nuint", StringComparison.Ordinal))
+                    if (typeNameBacking is "nuint" or "UIntPtr")
```
Exanite, line 2590: This change and the 3 other related changes are inconsistencies I noticed while working on the main issue. These are otherwise not related to the main issue addressed by this PR.

`typeNameBacking` eventually calls into `GetTypeName`, which can return either the keyword or the underlying type name depending on the generator configuration, so I believe this change is correct.

Exanite:
Going to mark ready for review for some feedback, but not ready for merge.

The new [CLongDefinesRegressionTestUnix](github.com/dotnet/ClangSharp/pull/699/commits/6c16210cafff781761e671c76598560b3606693d) test case represents a fairly harmless, but non-ideal case that I'm not quite sure how to handle.

The difference is the extra/redundant `unchecked` scope in `if (b > unchecked(18446744073709551615U) - a)` in the C# output. This is because `a` has `nuint` as its `targetTypeName`, which makes the generator believe that the integer literal to the left is also a `nuint` literal, but in reality is a `ulong` literal.

I'll keep investigating on my end, but my current ideas go in a few directions:
1. We make the generator understand widening/implicit casting rules. This is probably the most correct approach, but requires the most work, especially because I'm not familiar enough with these rules (especially on the C/cross platform side) to fully put them into code.
2. Accept the extra unchecked scope. It's effectively harmless, but not ideal.
3. Restrict my changes to apply to just `VarDecl`s. This is a bandaid fix, but might be reasonable since I can't think of cases where the generator won't output `unchecked` when it is actually needed. This will be much easier to investigate when an actual case is discovered.

`SDL_size_add_check_overflow`, the utility method the test case is based on, is notably something that shouldn't really be part of Silk 3's bindings either, so the benefit to fixing this isn't high, at least for Silk 3. I'm unfamiliar with how much ClangSharpPInvokeGenerator is used for porting code.

The diffs are otherwise exactly as I want them to be.
Tested with OpenCL, GL, AL, XR, Vulkan, and SDL through Silk 3's generator on both Windows and Linux.

tannergooding (dismissed):

tannergooding:
@Exanite there's some conflicts because I did a refactoring to split apart the large files into something a little more manageable.

Since its not a trivial fix from the web editor, could you resolve and push? Ideally that gets the tests passing as well

Exanite:
Sure thing, I can start looking into this some time tomorrow.

tests/ClangSharp.PInvokeGenerator.UnitTests/CTest.cs
```diff
@@ -1002,4 +1002,213 @@ readonly get
...
+";
+
+        var expectedOutputContents = @"namespace ClangSharp.Test
+{
+    public static unsafe partial class Methods
+    {
+        [return: NativeTypeName(""_Bool"")]
+        public static bool SDL_size_add_check_overflow([NativeTypeName(""size_t"")] nuint a, [NativeTypeName(""size_t"")] nuint b, [NativeTypeName(""size_t *"")] nuint* ret)
+        {
+            if (b > (18446744073709551615U) - a)
```
Exanite: This is the test that's currently failing.

This currently outputs as `unchecked(18446744073709551615U)`, which is non-ideal as I mentioned in my (non-review) comment, but depending on how you look at it, the behavior can be considered as being expected or harmless.

From my current understanding, this issue was here before I made my changes, but was hidden since IsUnchecked used to return false for this case.

I intentionally added this test to call out the behavior change and left it unfixed because the fix seems out of scope for this PR.

Please let me know how you want to move forward on this.

Exanite (commented):
Updated with changes from main.

tests/ClangSharp.PInvokeGenerator.UnitTests/CTest.cs
tannergooding: If you could open an explicit issue tracking it, I can take a look and we can just update the test to accept it for now

tannergooding (approved):

Exanite:
Thanks for the review and merge!

Issue #586, opened 2024-12-13
Ubuntu libclang still not resolved


GideonUng:
What dlls and to where. how do i get libclang.so files from Nuget even? the links to Nuget dont have any so files in them. what even are the limitations of dotnet that this isnt possible.

is it even possible to run this tool under linux currently. this has been an issue for years and there is neither a fix nor any semblance of documentation regarding this.

*****IMPORTANT*****
Failed to resolve libClangSharp.
If you are running as a dotnet tool, you may need to manually copy the appropriate DLLs from NuGet due to limitations in the dotnet tool support. Please see https://github.com/dotnet/clangsharp for more details.
*****IMPORTANT*****

Unhandled exception: System.DllNotFoundException: Unable to load shared library 'libClangSharp' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable: 
/home/gidcheen/.dotnet/tools/.store/clangsharppinvokegenerator/18.1.0.2/clangsharppinvokegenerator/18.1.0.2/tools/net8.0/any/libClangSharp.so: cannot open shared object file: No such file or directory
/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.10/libClangSharp.so: cannot open shared object file: No such file or directory
/home/gidcheen/.dotnet/tools/.store/clangsharppinvokegenerator/18.1.0.2/clangsharppinvokegenerator/18.1.0.2/tools/net8.0/any/liblibClangSharp.so: cannot open shared object file: No such file or directory
/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.10/liblibClangSharp.so: cannot open shared object file: No such file or directory
/home/gidcheen/.dotnet/tools/.store/clangsharppinvokegenerator/18.1.0.2/clangsharppinvokegenerator/18.1.0.2/tools/net8.0/any/libClangSharp: cannot open shared object file: No such file or directory
/usr/lib/dotnet/shared/Microsoft.NETCore.App/8.0.10/libClangSharp: cannot open shared object file: No such file or directory
/home/gidcheen/.dotnet/tools/.store/clangsharppinvokegenerator/18.1.0.2/clangsharppinvokegenerator/18.1.0.2/tools/net8.0/any/liblibClangSharp: cannot open shared object file: No such file or directory
[...]

PR #700, opened 2026-07-02
Fix native library resolution for dotnet tools on Linux

Happypig375:
## Problem

ClangSharpPInvokeGenerator fails to load `libclang` and `libClangSharp` on Linux when run as a dotnet tool, even though the RID-specific packages (e.g. `ClangSharpPInvokeGenerator.linux-x64`) bundle these libraries alongside the executable.

## Root causes

**1. Broken `TryResolveClang` method:** The custom `DllImportResolver` tried hardcoded versioned SONAMEs (`libclang.so.20`, `libclang-20`) that don't match the LLVM version ClangSharp is built against (21). This could find a wrong-version system library or fail entirely.

**2. Default resolver can't find co-located libraries:** The `SafeDirectories` search path should include the assembly's directory, but this doesn't work reliably for dotnet tools — especially AOT-compiled executables run from the NuGet cache via `dotnet tool run`. Both `libclang.so` and `libClangSharp.so` are co-located with the executable but never found.

## Fix

**1. Remove `TryResolveClang` entirely.** The versioned SONAME approach was inherently fragile (breaks with every LLVM major version bump, different distros use different naming conventions). The method was also unnecessary — when it returned `IntPtr.Zero`, the default resolver should have handled it.

**2. Explicitly search the assembly's own directory.** After the user-provided `ResolveLibrary` event, try loading from `Path.GetDirectoryName(assembly.Location)`. This handles all native dependencies (`libclang`, `libClangSharp`) uniformly on all platforms, without needing versioned SONAMEs or `LD_LIBRARY_PATH` workarounds.

```csharp
private static IntPtr OnDllImport(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
{
    if (TryResolveLibrary(libraryName, assembly, searchPath, out var nativeLibrary))
        return nativeLibrary;

    var assemblyDir = Path.GetDirectoryName(assembly.Location);
    if (assemblyDir is not null && NativeLibrary.TryLoad(Path.Combine(assemblyDir, libraryName), out nativeLibrary))
        return nativeLibrary;

    return IntPtr.Zero;
}
```

This is a follow-up to #586 which was closed as resolved by the RID-specific tool package support (20.1.2.2+). While the RID package mechanism correctly makes native binaries available alongside the executable, the custom resolver prevented the default resolution path from finding them on Linux.


tannergooding:
Will get to looking at this (and other PRs) tomorrow. Sorry for the delay, I've been fairly heads down with everything going on.

sources/ClangSharp.Interop/clang.cs
```diff
@@ -29,27 +30,24 @@ private static IntPtr OnDllImport(string libraryName, Assembly assembly, DllImpo
...
+
+        if (NativeLibrary.TryLoad(Path.Combine(AppContext.BaseDirectory, libraryName + suffix), out nativeLibrary))
         {
             return nativeLibrary;
         }
 
         return IntPtr.Zero;
     }
 
-    private static bool TryResolveClang(Assembly assembly, DllImportSearchPath? searchPath, out IntPtr nativeLibrary)
```
tannergooding, line 40: We explicitly want to keep this functionality.

I'm fine with explicitly trying the application directory first on non-Windows (which then is similar to `SafeDirectories` for Windows which is `application directory`, `%windir%\system32`, and `user directories in the dll search path`)

But we do want to search the system folder still otherwise and check for the various names it can bind to there

tannergooding: Since `LoadLibraryHelper` on unix ignores the flags, we'd rather want to effectively do something like
```
if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
    if (searchPath.GetValueOrDefault().HasFlag(DllImportSearchPath.ApplicationDirectory))
    {
        // Try the names using AppContext.BaseDirectory
    }

    // Try the standard variations
}
```

For apple target we want to try just the name + .dylib

For Linux we want to try the 3 variations already covered, in that order; first for app directory then for the normal system search

sources/ClangSharp.Interop/clang.cs
```diff
@@ -29,21 +30,54 @@ private static IntPtr OnDllImport(string libraryName, Assembly assembly, DllImpo
...
+            {
+                return nativeLibrary;
+            }
         }
 
         return IntPtr.Zero;
     }
 
-    private static bool TryResolveClang(Assembly assembly, DllImportSearchPath? searchPath, out IntPtr nativeLibrary)
+    private static bool TryResolveClang(string libraryName, string? baseDirectory, out IntPtr nativeLibrary)
```
tannergooding: nit, just make `baseDirectory` non-nullable and pass in the empty string, `""`

Then you can just `Path.Combine(baseDirectory, libraryName)` which already optimizes for one of them being empty.

---------

Actually, thinking further, while this code is then "technically correct" given the current state of the repo, this code is more general purpose and somewhat shared across different repos I maintain, so the general setup I have is fairly intentional.

The full change here really should be something like:
```csharp
private static IntPtr OnDllImport(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
{
    // Let user hooks attempt to resolve first, so they can override the behavior for their app

    if (TryResolveLibrary(libraryName, assembly, searchPath, out var nativeLibrary))
    {
        return nativeLibrary;
    }

    // Then try the normal resolution of the library

    if (TryResolve(libraryName, assembly, searchPath.GetValueOrDefault(), out nativeLibrary))
    {
        return nativeLibrary;
    }

    // Finally, do library specific resolution as applicable

    if (libraryName.Equals("libclang", StringComparison.Ordinal))
    {
        if (TryResolveClang(assembly, searchPath, out nativeLibrary))
        {
            return nativeLibrary;
        }
    }

    return default;
}

private static bool TryResolve(string libraryName, Assembly assembly, DllImportSearchPath? searchPath, out IntPtr nativeLibrary)
{
    if (TryResolveFromAppDirectory(libraryName, assembly, searchPath.GetValueOrDefault(), out nativeLibrary))
    {
        return nativeLibrary;
    }

    return NativeLibrary.TryLoad(libraryName, assembly, searchPath, out nativeLibrary)
}

private static bool TryResolveClang(Assembly assembly, DllImportSearchPath? searchPath, out IntPtr nativeLibrary)
{
    if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
    {
        return TryResolve("libclang.so.21", assembly, searchPath, out nativeLibrary)
            || TryResolve("libclang-21", assembly, searchPath, out nativeLibrary)
            || TryResolve("libclang.so.1", assembly, searchPath, out nativeLibrary);
    }

    nativeLibrary = default;
    return false;
}

private static bool TryResolveFromAppDirectory(string libraryName, Assembly assembly, DllImportSearchPath searchPath, out IntPtr nativeLibrary)
{
    nativeLibrary = default;

    // DllImportSearchPath is ignored on non-Windows and the default resolution algorithm does not look in the application directory
    // we need to look here for any dependencies that are shipped SxS with the application, using the standard variations allowed

    if (OperatingSystem.IsWindows())
    {
        return false;
    }

    if (((searchPath & (DllImportSearchPath.SafeDirectories | DllImportSearchPath.ApplicationDirectory)) == 0))
    {
        return false;
    }

    if (Path.IsPathFullyQualified(libraryName))
    {
        return false;
    }

    // We use Environment.ProcessPath because that is the location of the executable that launched the process. AppContext.BaseDirectory
    // can be overriden and is rather more the "AssemblyDirectory". This is a subtle distinction, but an important one.

    var applicationDirectory = Path.GetDirectoryName(Environment.ProcessPath);
    var libraryPath = Path.Combine(applicationDirectory, libraryName);

    if (NativeLibrary.TryLoad(libraryPath, assembly, searchPath, out nativeLibrary))
    {
        return true;
    }

    // We try the given name first, as that is always preferred. Then we try, in order:
    // * libname
    // * name.ext
    // * libname.ext
    //
    // This can result in some redundant lookups, but its how the built-in logic works for non-relat

    var prefixedLibraryName = "";

    if (!libraryName.Contains(Path.DirectorySeparatorChar))
    {
        prefixedLibraryName = $"lib{libraryName}";
        libraryPath = Path.Combine(applicationDirectory, prefixedLibraryName);

        if (NativeLibrary.TryLoad(libraryPath, assembly, searchPath, out nativeLibrary))
        {
            return true;
        }
    }

    var extension = ".so";

    if (OperatingSystem.IsMacOS() || OperatingSystem.IsIOS() || OperatingSystem.IsTvOS())
    {
        extension = ".dylib";
    }

    libraryPath = Path.Combine(applicationDirectory, $"{libraryName}{extension}");

    if (NativeLibrary.TryLoad(libraryPath, assembly, searchPath, out nativeLibrary))
    {
        return true;
    }

    if (prefixedLibraryName.Length != 0)
    {
        libraryPath = Path.Combine(applicationDirectory, $"{prefixedLibraryName}{extension}");

        if (NativeLibrary.TryLoad(libraryPath, assembly, searchPath, out nativeLibrary))
        {
            return true;
        }
    }

    return false;
}
```

Happypig375: Are you really going to hardcode 21 there (the 20 not being updated to 21 caused loading issues already) or should it be determined somehow?

tannergooding: The general issue is not being able to trivially validate it was actually updated and is "correct". This is what is doing the load for `libClang` so we can't just check or use `clang.getClangVersion`, for example.

I guess we could have some `int MajorVersion => 21; int MinorVersion => 1` exposed off `@clang` here and have that feed into the lookup logic. Then try to catch future errors with a test that validates it against what clang.getClangVersion` returns. That can then also replace the `ExpectedClang/ClangSharpVersion` that the pinvoke generator validates against.

It's all around problematic though, just because its a bit recursive and so there's no trivial way to guarantee it was in fact updated properly. You may just fail to run at all because it fails to resolve again in some contexts.

Happypig375:
I also doubt the correctness of .dylib for iOS and tvOS. Apple requires dynamically linked libraries to be made as frameworks instead of raw dylibs, or they will not pass App Store review.

However, currently clang does not build for iOS and tvOS - it will be relevant when those native targets get made.

tannergooding (approved):

tannergooding:
> I also doubt the correctness of .dylib for iOS and tvOS. Apple requires dynamically linked libraries to be made as frameworks instead of raw dylibs, or they will not pass App Store review.

This just matches what the .NET library loader logic does. Its a matter of consistency with the underlying framework.

PR #747, opened 2026-07-13
Add a log-invalid-remappings switch for ABI-illegal remappings

tannergooding:
This adds a new opt-in `--log-invalid-remappings` config switch that emits a `DiagnosticLevel.Info` diagnostic for each configured remapping whose value is structurally not a legal unmanaged type name, scoped precisely to the array/fixed-buffer shape (a numeric subscript `[<digit>`, the same rule as the lowering, so a managed `int[]` with an empty subscript is left untouched).

The message names the remapping and states the applied lowering, e.g.:

> Potential invalid remapping 'MyBuffer=sbyte[8]'. A fixed-size-buffer type is not directly expressible in an unmanaged signature; it was lowered to the pointer 'sbyte*'.

The lowering used to compute the shown pointer is factored into a shared `TryLowerFixedSizeBufferRemapping` helper so the diagnostic and `GetRemappedTypeName` can't drift.

----------

This builds on #745 (the fixed-size-buffer-to-pointer lowering), which has since merged; the branch is based directly on that commit so the diff here is just the diagnostic feature.

### DIFF
--- a/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.Naming.cs
+++ b/sources/ClangSharp.PInvokeGenerator/PInvokeGenerator.Naming.cs
@@ -812,6 +812,16 @@ public sealed partial class PInvokeGenerator
             nativeTypeName = string.Empty;
         }
 
+        if (TryLowerFixedSizeBufferRemapping(remappedName, out var loweredName))
+        {
+            remappedName = loweredName;
+        }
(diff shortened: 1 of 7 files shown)


tannergooding:
Dropping this — not high enough value relative to the number of ABI edge cases it would need to cover to be trustworthy.

PR #781, opened 2026-07-15
Add support for scraping struct over-alignment as NativeAlignment

tannergooding:
Structs declared with `__declspec(align(N))` / `DECLSPEC_ALIGN` (e.g. `CONTEXT`, `SLIST_ENTRY`: align 16 on x64/arm64) lost their requested alignment in the generated bindings. .NET's `StructLayout.Pack` is `min(specified, natural)` and can only *lower* alignment, never raise it, so ClangSharp previously dropped over-alignment entirely — it only captured under-alignment via `pack = alignment < maxAlignm ? alignment : null`. Downstream consumers (see microsoft/win32metadata#1044) then had nothing to preserve, resulting in `ERROR_NOACCESS` from `GetThreadContext`/`SetThreadContext`.

This scrapes the requested alignment and surfaces it so the information isn't silently lost.

----------

**Detection.** Over-alignment is `requestedAlignment > maxAlignm`, where `requestedAlignment = recordDecl.MaxAlignment / 8` (Clang's `Decl::getMaxAlignment()`, in bits, already exposed through `Decl.MaxAlignment` / `CXCursor.MaxAlignment`) and `maxAlignm` is the existing max natural field alignment. No native rebuild required — this reuses the existing interop plumbing.

**Warning (always on).** Whenever a struct is over-aligned, a warning is emitted by default stating that .NET cannot honor it and what alignment the runtime will actually use, e.g.:

> Struct 'MyStruct' requests 16 byte alignment which .NET cannot honor; over-alignment can only lower, never raise, alignment, so the runtime will align it to 4 bytes.

**Attribute (opt-in).** A new `generate-native-alignment-attribute` option (`GenerateNativeAlignmentAttribute`, `1L << 44`) gates the rest, consistent with `generate-native-bitfield-attribute` / `generate-native-inheritance-attribute`. When enabled, an over-aligned struct emits:

- `[NativeAlignment(N)]` on the struct, plus the generated `NativeAlignmentAttribute.cs` helper (`[Conditional("DEBUG")]`, `int` ctor — matching the other documentary attributes, which win32metadata already reads from the generated C# source via Roslyn rather than compiled metadata).
- The requested alignment encoded into `Pack` (`[StructLayout(LayoutKind.Sequential, Pack = N)]`) even though .NET clamps it, so the metadata round-trips correctly for win32metadata / Rust / other consumers. `Pack` is only auto-set when a `#pragma pack` value isn't already present.
- XML output gains `alignment="N"` on `<struct>`.

Example (flag enabled) for `struct __attribute__((aligned(16))) MyStruct { int x; };`:

```csharp
[StructLayout(LayoutKind.Sequential, Pack = 16)]
[NativeAlignment(16)]
{
    public int x;
}
```

----------

**Known limitations (intentionally out of scope):**

- **Per-architecture layout.** This does not resolve the per-architecture field/layout differences of `CONTEXT`/`SLIST_ENTRY`. The attribute is captured per-arch and win32metadata's existing one-pass-per-arch merge preserves it, which is sufficient for that consumer.
- **Empty over-aligned structs.** Detection uses `maxAlignm` (max field alignment) as the natural baseline; a field-less struct defaults `maxAlignm` to the over-aligned alignment, so an empty over-aligned struct neither warns nor emits the attribute. The target types (`CONTEXT`, `SLIST_ENTRY`) have fields, so they're covered.

----------

**Tests.** Added golden-file regression tests to `StructDeclarationTest.cs` — `OverAlignmentTest` (flag off: warning only, no attribute/`Pack`) and `OverAlignmentWithNativeAlignmentAttributeTest` (flag on), sharing the expected diagnostic and using portable `__attribute__((aligned(16)))` over a naturally-4 struct so output is uniform across the OS axis. Baselines were generated for all 16 variants and unified to a single `{Case}.{Mode}` file each. Full solution builds with 0 warnings; the generator suite passes (3706 generator tests).


jevansaks:
Woohoo! When can we expect an updated nuget with this?

tannergooding:
Its already up :)

PR #787, opened 2026-07-15
Add glob matching to the name-matching options

tannergooding:
Follow-up to #784, which unified the CLI but left the name matcher exact-match-only despite the option descriptions advertising `Supports wildcards.`

The `--with-*` value options, their paired `--without-*` opt-outs, and the `-i`/`-e` include/exclude lists now do genuine shell-style glob matching, so patterns like `Foo*`, `*Bar`, `NS::*`, and interior `*`/`?` match declaration names. The bare `*` catch-all keeps its existing all-or-nothing semantics.

----------

### Syntax
- `*` matches any run of characters (including qualification separators); `?` matches a single character. Every other character is a literal.
- Matching is case-sensitive and treats `::` and `.` as equivalent separators, consistent with `QualifiedNameComparer`, so `NS::*` and `NS.*` behave the same.

### Precedence
Exact match wins over any glob; among globs the most-specific (most literal characters) wins, with ties broken by registration order; the bare `*` catch-all remains the lowest-priority fallback. Opt-outs (`--without-*`, `-e`) support globs too, symmetric with the opt-ins; an exact opt-in still beats a glob opt-out.

### Performance
The hot path is unchanged: an exact hash lookup is tried first and the precompiled globs are only iterated on a miss. Patterns are compiled to anchored regexes once per store and cached, so non-glob configurations pay effectively nothing.

----------

### Scope
Globbing covers the options routed through the central lookup helpers plus include/exclude and manual-imports. `--with-class`/`--with-namespace` keep their existing bespoke trailing-`*` prefix matching, and `--with-transparent-struct` stays exact-only (both are too entangled in type resolution to fold in here); their descriptions were made accurate rather than left claiming full glob support. These can be unified in a follow-up if desired.

### Tests
Golden-file regressions covering prefix/suffix/`?` excludes, the `?` single-character wildcard, the full exact > glob > bare-`*` precedence chain (via calling conventions, which render distinctly), and glob opt-outs. Baselines were regenerated with `UPDATE_BASELINES=1` and unified across the OS/config matrix.

Local `dotnet build -c Release` is clean (0 warnings) and the full `dotnet test -c Release` suite is green.


tannergooding:
> [!NOTE]
> The remaining CI failures on this PR are unrelated to globbing: the 9 \GeneratedCode\/\HelperTypes\/\TransparentStruct\ baselines assert the stale \21.1.8.3\ version string after the \VersionPrefix\ bump to \21.1.8.4\ didn't regenerate them (already red on \main\). Fixed separately in #789 — once that merges I'll rebase this PR and CI should go green. This PR's own build is 0-warning and all glob tests pass.

